PR #38, opened 2016-02-09
Change Spring Evals Vote From Secret Immediate to Immediate

liam-middlebrook:
Instead of doing a secret vote during membership evaluations during Spring
Evals, a non-secret immediate vote will occur.


liam-middlebrook:
Vote Failed

PR #40, opened 2016-02-15
Changes to Executive Board elections

bmbowdish:
Executive Board uses alternative vote now.

Created Alternative Vote as a voting method and applied it to E Board
elections.


bmbowdish:
I swear to god.

PR #43, opened 2016-02-26
Added House Meeting in ByLaws

bmbowdish:
Located in section 2.B in ByLaws


liam-middlebrook:
Reminder to bring the ballots for this to an upcoming house meeting so it can be discussed and then voted in.

PR #48, opened 2016-04-04
Voluntary Attendance for Discussions of Semantic Changes

bmbowdish:
==
Main Amendment
==
This allows members who are disinterested in the amendment process or have already made their mind a way to abstain from discussion and the vote. 
Altered the section section that defines the amendment process to allow this change. 

==
Comments
==
This is Max's Amendment. 
Discussion will be on April 10th, 2016. 
I'd appreciate help from Matt to make sure this is worded correctly.


articles.tex
```diff
@@ -99,7 +99,9 @@
 \asubsection{Semantic Changes}
 Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
 Any modifications made due to the discussion are added to the written proposal and the modified proposal is posted in the House during the week.
-The final proposal is presented the following week and ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
+The final proposal is presented the following week at the end of House Meeting.
```
gambogi: I would leave out the "end of" line. The chair of the meeting will figure out what time in the meeting makes sense.

articles.tex
```diff
@@ -99,7 +99,9 @@
...
+Attendance for the discussion on the change is not required, but only members who stay until the end of discussion will receive ballots.
```
gambogi: How does this intended to interact with quorum? 
My reading of this makes it seem like quorum is counted at the end of discussion, which would bring back a problem we had a few years ago. If there are a handful of members who object to the premise of a change, they can skip the discussion and make it impossible to hold a vote.

Maxbgenius:
Can we change the title to something more neutral like "Make Discussion on Amendments Voluntary"?

gambogi:
Does that seem like a fair description?

articles.tex
Maxbgenius: The idea is that anyone who chooses to leave before discussion would be choosing to abstain from all amendments being voted on that night.

gambogi: I think there needs to be some language in this amendment that specifies that. Right now I think it will reintroduce an old problem.

Maxbgenius:
The discussion must be held but members are not required to stay for it if they choose to abstain. how about "Make Attendance at Discussion of Semantic Changes Voluntary"

Maxbgenius:
That works for me

liam-middlebrook:
Could you rebase these commits so they have a more descriptive title for what the goal of this change is? That way if this is accepted and merged people will be able to understand the change by reading the git commit log?

gambogi:
@wastevensv Looks like that was preexisting but now's a perfectly fine time to fix it.

liam-middlebrook:
FAILED 10 to 36

liam-middlebrook:
Edited proposal as of https://github.com/ComputerScienceHouse/Constitution/pull/48/commits/9732fd603116f90e96d34746482ab746273b117e

articles.tex
```diff
@@ -97,11 +97,10 @@
 A quorum of fifty percent of the Total Number of Possible Votes is required for passage.
 
 \asubsection{Semantic Changes}
-Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
-Any modifications made due to the discussion are added to the written proposal and the modified proposal is posted in the House during the week.
-The final proposal is presented the following week at the end of House Meeting.
-Attendance for the discussion on the change is not required, but only members who stay until the end of discussion will receive ballots.
-At the end of discussion, ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
+Any semantic change to the Constitution requires the change to be brought up at House Meeting and proposed in writing for discussion at the following House Meeting.
+The text of the change may be modified up until the aforementioned House Meeting.
```
jeid64: Which aforementioned house meeting? You mention two. No idea which one is referred to. Why the use of the word aforementioned anyway?

articles.tex
```diff
@@ -97,11 +97,10 @@
...
 \asubsection{Semantic Changes}
-Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
-Any modifications made due to the discussion are added to the written proposal and the modified proposal is posted in the House during the week.
-The final proposal is presented the following week at the end of House Meeting.
-Attendance for the discussion on the change is not required, but only members who stay until the end of discussion will receive ballots.
-At the end of discussion, ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
+Any semantic change to the Constitution requires the change to be brought up at House Meeting and proposed in writing for discussion at the following House Meeting.
+The text of the change may be modified up until the aforementioned House Meeting. 
+At the beginning of the discussion period, ballots are distributed for a balloted House vote as described in \ref{Balloted Vote}.
+Attendance for the discussion on the change is not mandatory.
```
jeid64: This is still really vague. No clue what this does for quorum. If attendance is not mandatory, do I still get a ballot? What happens if down the line someone reads this and thinks it means that "attendance for the discussion on the change is not mandatory" means "it is up to the chair to host discussion as it is optional." 

Might want to mention that discussion for changes on amendments happen at the end of house meeting. Not really necessary though. 

Also, probably want to refer to it as the "proposed amendment" instead of "the change"

Maxbgenius: Thanks for the input, let me change a few things and see if that addresses most of your concern.

Maxbgenius:
I am probably doing this comment thing wrong but hopefully the message is conveyed nonetheless. 

I changed the "aforementioned" thing to directly reference the house meeting where the change is being discussed. 

I refer to it as "the change" instead of the "proposed amendment" only because that's how it was written before. I'd guess they wrote it that way to match the section title (semantic changes).

I think I can trust future members to understand that "attendance for the discussion on the change is not mandatory" refers to attendance at the discussion, not the discussion itself.

PR #55, opened 2016-04-17
Executive Board Confidentiality and Safe Environment

speakerbug:
Check one:

Summary of change(s):

Adds keeping confidentiality as an Executive Board responsibility when a House member requests for something to be kept private, except in cases where there is a potential for harm to themselves or others. Also, adds a responsibility for Executive Board to proactively maintain a safe environment for its members

To be announced on April 17th, 2016. To be voted on April 24, 2016.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -303,7 +303,7 @@ However, the Chairman and at least two-thirds of the Voting Members must be pres
 	\item To act as a Judicial Board as defined in \ref{Judicial}.
 	\item To review major projects, as defined in \ref{Expectations of House Members},  presented to them by the Evaluations director.
 	\item To make the final vote regarding conditionals and appeals as defined in \ref{Membership Evaluation}.
-	\item To respect the privacy of House members confiding in the Executive Board, barring situations related to endangerment of oneself or others or in the case of a Judicial Board.
+	\item To respect the privacy of House members confiding in the Executive Board, barring situations related to endangerment of oneself or others, sexual assault, or in the case of a Judicial Board.
 	\item To publish a document at the end of each semester to all Members stating House’s accomplishments of that semester.
 	\item To review and update the Constitution at the end of each standard operating session, as defined in \ref{Standard Operating Session}.The constitution should remain up to date with current practices.
 \end{enumerate}
--- a/articles.tex
+++ b/articles.tex
@@ -303,6 +303,8 @@ However, the Chairman and at least two-thirds of the Voting Members must be pres
 	\item To act as a Judicial Board as defined in \ref{Judicial}.
 	\item To review major projects, as defined in \ref{Expectations of House Members},  presented to them by the Evaluations director.
 	\item To make the final vote regarding conditionals and appeals as defined in \ref{Membership Evaluation}.
+	\item To proactively maintain a safe and welcoming enviornment to House members.
+	\item To respect the privacy of House members who confide in confidence barring circumstances that allude endangerment of ones self or others.
 	\item To publish a document at the end of each semester to all Members stating House’s accomplishments of that semester.
 	\item To review and update the Constitution at the end of each standard operating session, as defined in \ref{Standard Operating Session}.The constitution should remain up to date with current practices.
 \end{enumerate}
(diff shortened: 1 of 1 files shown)


harlanhaskins:
I think the wording of 'confide in confidence' is a little funny, but other than that I'm 👍

speakerbug:
I wasn't sure about that either, but the exact definition of confide is "tell someone about a secret or private matter while trusting them not to repeat it to others" and basically members will be doing that and be doing so in confidence. Feel free to suggest other ways to word it though.

articles.tex
```diff
@@ -303,6 +303,8 @@
 	\item To act as a Judicial Board as defined in \ref{Judicial}.
 	\item To review major projects, as defined in \ref{Expectations of House Members},  presented to them by the Evaluations director.
 	\item To make the final vote regarding conditionals and appeals as defined in \ref{Membership Evaluation}.
+	\item To proactively maintain a safe and welcoming enviornment to House members.
```
nickdepinet: Enviorment->environment

articles.tex
```diff
@@ -303,6 +303,8 @@
...
+	\item To respect the privacy of House members who confide in confidence barring circumstances that allude endangerment of ones self or others.
```
ajgajg1134: I think this is much easier stated "To respect the privacy of House members who confide in them" I don't think it's necessary to provide a specific opposite case as it can be considered implied (Not to mention any E-Board members who may be mandatory reporters under Title IX )

harlanhaskins: I'm not too sure it's considered implied. I can absolutely see a future discussion where this passage comes up. I'd prefer explicitly stating this.

JDrit:
I fail to see the point of this change. The first bullet point is so vague that it can be applied to almost anything and is for all intensive purposes useless.

speakerbug:
I agree that the first bullet point is vague, but my intent is to make sure that this is known to eboard members that this a responsibility of them and that they should be actively trying to ensure this is happening instead of just doing cleanup after major events happen.

JDrit:
This is a quote of SIH Constitution: "...ensure each SIH sustains an environment that is clean, livable, and welcoming to all people."

This amendment does not change anything since we derive our power from the SIH Constitution and thus this is just amendment is just duplicated text.

speakerbug:
I personally don't mind having the text duplicated since I don't think most people read the SIH constitution but I know eboard members read ours and I don't see the harm in ensuring it's very clear.

harlanhaskins:
Also, @JDrit could you point me to the current, up-to-date SIH constitution?

JDrit:
Having text duplicated just causes bloat among all of our documents. 

Also as we have learned time and time again in the past, we can't have vague sections in our constitution. This only leads to miscommunication among members and people not understanding what needs to happen. 

If you want to add additional requirements to E-Board members, then you need to make sure that everything is clearly spelled out and understood by all. That is required to be fixed before we even begin to think about this change.

fwi: SIH Constitution: https://wiki.csh.rit.edu/wiki/SIH_Constitution

speakerbug:
I just updated the text to remove the duplicate text from the SIH constitution. This is more of a cultural change than a amendment anyway. I also added an exception for confidentiality to be spoken about during a Judicial Board if it seems fit.

articles.tex
```diff
@@ -303,6 +303,7 @@
 	\item To act as a Judicial Board as defined in \ref{Judicial}.
 	\item To review major projects, as defined in \ref{Expectations of House Members},  presented to them by the Evaluations director.
 	\item To make the final vote regarding conditionals and appeals as defined in \ref{Membership Evaluation}.
+	\item To respect the privacy of House members who confide in confidence barring circumstances that allude endangerment of ones self or others or in the case of a Judicial Board.
```
ryansb: Changing to:

> To respect the privacy of House members confiding in the Executive Board, barring situations related to endangerment of oneself or others...

Does that wording sit better with everyone?

speakerbug: I think that's a fine rewording. What are other peoples thoughts?

nickdepinet: Ryan's rewording reads a lot better.

liam-middlebrook:
Please remember to recompile the constitution before your PR is merged (assuming it is approved by House).

liam-middlebrook:
PASSED 33-2

PR #60, opened 2016-05-02
Dues Amount

mbillow:
Check one:

Summary of change(s):
Every member has the same right to every house resource. There is no reason to charge different dues amounts, it only complicates collection.


liam-middlebrook:
Passed 44-2

PR #62, opened 2016-05-23
Dues Amount and Dues Collection Period

mbillow:
Check one:

Summary of change(s):
Amendment One: All dues (regardless of on/off-floor status are the same.
Amendment Two: Dues can be collected at any point in the year, just not after Spring (Membership) Evaluations. Dues collection at the beginning of the year is defined by the Financial Direction in conjunction with RIT Administration (CRL and Housing).


mbillow:
Passed 44-2

PR #64, opened 2016-06-11
Add docs

ajgajg1134:
Just adding a simple contribution guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,15 @@ The master branch's HEAD is the current legal constitution of CSH.
 
 You can demo changes etc on forks or branches though. In fact, the diff of this branch/fork and the head of constitution/master would make it extremely easy to illustrate the changes you've made relative to the current constitution, or for people to annotate your changes with comments, etc.
 
+# Contributing to the CSH Constitution
+
+## Requirements
+Must have make, and pdflatex installed
+
+## Step by Step
+
+1. Fork this repo
+2. Make a new branch on your fork
+3. Modify the `.tex` files with your changes
+4. Commit, Push and make a Pull Request
+5. Wait for vote to pass then merge changes


liam-middlebrook:
I don't see any compelling reason to add this guide. There are already instructions in the document itself (articles.tex) that define the process. This only adds another location that would need to be changed if the process was altered. 

In addition why make a separate file from README.md? The README is short enough that anyone who is going to contribute should be able to read it with ease.

ajgajg1134:
I disagree in that the document itself merely defines the process by which we ratify amendments. Wheras the purpose of this PR is to make it easier for those who wish to modify this repository which is how we currently store the constitution. I agree the README is really small too, I only separated it in case people wanted to expand on it further and therefore not clutter the README, but I see your point.

I'm curious what others have to say, specifically constitutional maintainers.

JDrit:
I think that this would be useful l, especially for people who do not have an in depth knowledge of working with latex files. It does not have to be anything official so we can change the doc easily in the future. 

tl;dr I agree with Glaude

gambogi:
I say we add to the current readme until it starts to feel unwieldy. Otherwise I do agree there is a need for a bit more documentation of how to build the constitution.

CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Requirements
+Must have make, and pdflatex installed
+
+## Step by Step
+
+1. Fork this repo
+2. Make a new branch on your fork
+3. Modify the `.tex` files with your changes
+4. Run make *TWICE*
```
gambogi: Shouldn't have to run make twice anymore, the current makefile should compile twice so the links work. If that's not the case then we should make it the case.

liam-middlebrook:
@ajgajg1134  I'm happy to merge this if you concatenate it all into README.md

liam-middlebrook:
And squash these commits into one :smiley_cat: !

ajgajg1134:
Squashed and moved to README.md

PR #71, opened 2016-08-28
Fix pronoun ambiguity loopholes

liam-middlebrook:
Check one:


mbillow:
I would consider this a non-semantic change. The rules have always applied to all genders.

mbillow:
Also, we should make sure these are the only two instances of this, I feel like there are more.

liam-middlebrook:
I searched for ' his' case insensitive and didn't see any other occurrences.

mbillow:
There is also at least one instance of "he".

liam-middlebrook:
That should be fixed now.

bylaws.tex
```diff
@@ -253,9 +253,9 @@
 The Membership Evaluation is responsible for determining those individuals who will have the option to continue Active Membership in the following year.
 Semi-Annual Evaluations Meetings are open only to all current Active Members.
 All Executive Board members are expected to attend the Semi-Annual Evaluations Meetings to assist in the evaluations of House members.
-\\* \\*
-At the beginning of any Evaluation Process listed herein, the Evaluations Director must read the sections of the Computer Science House Constitution and By-Laws used during the respective Evaluation Process.
-It is incumbent upon each House member to provide the Evaluation Director with whatever information he or she feels is necessary to ensure an accurate evaluation.
+\v* \\*
+Anning of any Evaluation Process listed herein, the Evaluations Director must read the sections of the Computer Science House Constitution and By-Laws used during the respective Evaluation Process.
```
ajgajg1134: Is this change intentional?

PR #72, opened 2016-08-28
Process for Evals to Disable Accounts after Failure to Pass Evaluations or Resignations

liam-middlebrook:
Check one:

Summary of change(s):

Evals should notify OpComm about which account need to be disabled after evaluations meetings.


liam-middlebrook:
Also added in guidelines for this same process if intro or honorary members resign.

gambogi:
First, I prefer "guidelines" kept out of the constitution; however, if there is consensus that a line like this is needed for godsakes add it in one place in the responsibilities of Eval along the lines of: 
"To notify the Root Type Persons of any changes in standing as it pertains to account access control."

ehouse:
This is still pretty vague. Notify the RTP's within x days in days after failure is more to my taste. Same goes for failure of a conditional,

stevenmirabito:
@ehouse I agree, but it should be worded in a way so that there isn't a _maximum_ amount of time, so people don't technically get to keep their accounts if the evals director forgets to notify the OpComm director within the given amount of time.

gambogi:
I mean, technically this entire addition is unnecessary/already implied. Non-members don't have the right to an account. It's obviously the job of EBoard and the RTPs to coordinate on maintaining who does and does not have access. You can write up the scariest deadline language you like but I doubt any actual functional change will come of it. 

I know historically we have sometimes been slow to officially pull account access if the person was appealing or reapplying.

bmbowdish:
There is a nice line between blatantly telling the Evaluations Director to hand in the list of failures in X days, which is not something that fits into the constitution, and leaving it so open ended that it never gets done. 

I don't know if this amendment is entirely necessary, I mean think of all of the things in the constitution that we don't actively follow or care about, but having this process codified is certainly helpful for later down the line and allows us to say "Hey, evals director! You need to do this!"

This is just like the CoC amendment. Much like that amendment, this makes too much sense to vote no. Thank you for writing this Liam.

liam-middlebrook:
I believe it's much easier to write this intro the places that already include the conditions under which an account should be revoked. To add this to the responsibilities of the evals director would be to add what I consider superfluous and over-complicate the section mentioned above.

gambogi:
> I believe it's much easier to write this intro the places that already include the conditions under which an account should be revoked. 
How is a scattered four statement change easier?

> To add this to the responsibilities of the evals director would be to add what I consider superfluous and over-complicate the section mentioned above.
How would only adding one line to the responsibilities of eval be superfluous and over-complicating compared to adding the same line four times all over the place. 

Using the approach as written means that if we ever add another method of membership status change that affects account rights we would need to also remember to include this exact text all over the place. If we want to tweak the restrictions on this notification period we will have to remember all of the sections that this exact text is.

liam-middlebrook:
> > I believe it's much easier to write this intro the places that already include the conditions under which an account should be revoked.
>
> How is a scattered four statement change easier?
> 
>> To add this to the responsibilities of the evals director would be to add what I consider superfluous and over-complicate the section mentioned above.
>
> How would only adding one line to the responsibilities of eval be superfluous and over-complicating compared to adding the same line four times all over the place?
> 

I don't believe that the text that summarizes the different conditions under which a member can have their account access revoked fits well into a bulleted list entry. You need to account for failure to pass introductory evaluations, failure to pass the membership evaluation that you attend directly after aforementioned introductory evaluation, resignation of introductory members, and resignation of honorary members. Now if you believe that the previous sentence is effective please feel free to submit another amendment to do so. I don't think that's the best way to convey this process.

> Using the approach as written means that if we ever add another method of membership status change that affects account rights we would need to also remember to include this exact text all over the place. If we want to tweak the restrictions on this notification period we will have to remember all of the sections that this exact text is.

Exactly, I don't see any issue with the addition of this type of statement if that is something that will happen.

gambogi:
> I don't believe that the text that summarizes the different conditions under which a member can have their account access revoked fits well into a bulleted list entry. 

You don't need to include that in the bulleted list. The line I included in my first comment accounts for all possible situations and is perfectly clear.

nickdepinet:
I think we need to be in a state where we're looking at new semantic changes to the constitution and considering several things before even bringing them to house for discussion. Burnout is a very real thing and if we constantly bring up multiple amendments every week, people will just get pissed off when actually critical amendments happen.

With that in mind, I don't think this change is necessary. We don't necessarily need to concretely define a date by which to remove an account. We already link account ownership to membership, and give RTPs & Eboard the right to remove accounts in the Code of Conduct. If it's a problem that peoples accounts aren't getting revoked in a timely fashion, then we have an operational problem, not a legislative one. I think that adding the documentation to the wiki is a better place for these operational guidelines to keep the constitution from becoming unnecessarily unwieldy.

Proposing for closure.

liam-middlebrook:
Agreed

PR #73, opened 2016-08-28
Add clause to allow for non-intro eval repeats

liam-middlebrook:
Check one:

Summary of change(s):

Adds a optional vote for intro members who have already passed the Introductory Evaluation to skip the Introductory Evaluation in order to bypass requirements that are targeted towards getting familiar with CSH.


gambogi:
Way this kind of thing is usually done is by override. I can see how needing 87%ish support could be prohibitive though. That said, a few things need to be clarified here:

What kind of vote may be held?
What is the required quorum, passing threshold?
How is it brought up? Evals discretion? Motion followed by a second?

stevenmirabito:
Is this house vote? An eboard vote? How many people of either group are required to vote to pass? The type of vote and quorum requirement should be specified. If you want to be specific, you should also specify the method of voting, but I'm OK with leaving that up to interpretation in this case.

gambogi:
If there is a vote in the constitution that does not have a method, quorum, and pass threshold associated that is a bug. This is something you cannot leave up to interpretation.

bmbowdish:
I believe this intended to be voted on during a Majority Vote at an evaluations meeting where there is quorum, the same quorum required for any other Evaluations vote. The way it would work would be

Vote Off Floor -> (Vote On Floor) -> Vote to Pass Fall Evals, move to the next application.

bylaws.tex
```diff
@@ -207,6 +207,7 @@
 	\item The applicant submits an application to the Evaluations Director for review.
 	\item The applicant participates in an informal interview with exactly three current Active, Alumni, or Honorary Members.
 	\item The application materials are then reviewed at an Evaluations meeting and the members present vote on whether to accept the person as an introductory member.
+	\item If the applicant has passed an Introductory Evaluation in the past an optional Simple Majority vote may be held to allow the applicant to automatically pass the Introductory Evaluation. Any member in attendance may motion for this vote to occur.
```
gambogi: Strike "optional". It's redundant when paired with the "may".

liam-middlebrook: This has already gone to ballot.

gambogi: It's not a semantic change?

liam-middlebrook: Fair, I'll change that if it passes. That way it's not confusing for people when it's in-flight.

PR #76, opened 2016-09-09
Add fixed percentage requirements for the different types of signatur…

VictoriaWeaver:
Check one:

Summary of change(s):

…es on the Intro Packet to obtain an account

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -355,7 +355,12 @@ Root Type Persons have the authority to manage user accounts for House systems. 
 \renewcommand{\theenumi}{\alph{enumi}} % For this section, we want items to use letters
 \begin{enumerate}
 	\item Sign the Code of Conduct sheets pertaining to the responsible utilization of Computer Science House and Rochester Institute of Technology facilities.
-	\item Obtain greater than or equal to 60\% (rounded up to the nearest whole person) of signatures in the Introductory Packet or successfully complete Introductory Evaluations.
+	\item Successfully complete Introductory Evaluations or obtain greater than or equal to 60\% (rounded up to the nearest whole person) of each of the following types of signatures in the Introductory Packet:
+		\begin{itemize}
+			\item All Resident Active members and Executive Board members
+			\item All Introductory members residing on floor
+			\item Fifteen (15) Non-Resident and/or Alumni members (not including Non-Resident Executive Board members)
+		\end{itemize}
 \end{enumerate}
 
 \end{document}


liam-middlebrook:
So I've been asked a few times what would happen if this an #75 were to both be passed. In my interpretation of these changes the following would become the new packet/account requirements.


In order to pass packet the following signatures are required:

* All Resident Active Members

* All Executive Board Members

* 15 Non-Resident / Alumni / (technically not, but it'll still happen: Honorary) Members (not including off-floor eboard members)

In order to get an account the following requirements must be met:

* You must obtain 60% of the available signatures (rounded up to nearest whole person) for the following groups
  - Resident Active Members / Eboard Members

  - Non-Resident / Alumni / (technically not, but it'll still happen: Honorary) Members (not including off-floor eboard members)

  - Resident Introductory Members

So if both pass, packet will still need to collect freshman signatures, but they won't count towards the passing of packet, only account creation. Packet completion percentage will need to count the three groups listed directly above separately, and the signatures missed field in evals will only include from the groups listed in the first bulletpoint section above.

gambogi:
Going to go ahead and say that passing both as is would be an insane complication to the process since they don't really make sense together as written. These should just come up in separate weeks (and since this was only posted today they will anyway). Once we have the results from this week's vote @VictoriaWeaver can choose to adjust the language here to more accurately reflect the spirit of the change if necessary.

liam-middlebrook:
Just to note: @gambogi they haven't been brought to House Meetings. They're technically both at the same stage in the process (pre-proposal). We're roughly two weeks out from any changes either way.

I personally think that @ajgajg1134 should have both of these discusses in the same period and then either @mbillow or @VictoriaWeaver would pull their ballots based on the discussion, that way we don't get this weird over-complicated merge.

gambogi:
Sorry, mixed up the dates for #75 and #73, that makes sense.

VictoriaWeaver:
I'm proposing this as an alternative option to [#75](https://github.com/ComputerScienceHouse/Constitution/pull/75).  I don't intend for both of them to pass since they are very clearly contradictory.  If presented properly at house meeting, then it shouldn't be an issue- people can vote for option A, option B or neither.

zedoax:
Wouldn't it be simpler and achieve the same effect if the amendment were to require 15 Resident Intro Members, similar to the Non-Resident/Alumni requirement?  I feel that the 60% for three categories makes the packet requirements overly complex with more than one percentage to keep track of.

VictoriaWeaver:
@zedoax that gets tricky because the number of on-floor freshmen fluctuates each year.  There's a possibility that there could be a freshmen class with <15 on-floors, so there would have to be an override for that year.  With the Non-Residents/Alumni we don't run that risk.

As for complicating the packet, I actually think it would make things easier at Fall Evals.  When someone who didn't finish their packet comes up, one of the first questions is always "how many of their missed signatures were in X category?".  I view it as simply doing the math ahead of time.

zedoax:
@VictoriaWeaver 1) Would it then be possible to word it in a way that it would be 15 freshmen (or less if there are less freshmen applying than 15)
2) Instead of then requiring 60% from each category, how about implementing percentage trackers on Tal Packet that give you that breakdown to keep track of

VictoriaWeaver:
@zedoax (2) is already something I was planning on requesting if the amendment passes (but might be good either way as long as it's recorded in the evals db after the fact).  I'm not comfortable with (1) because it seems too unstable.   That stability is already provided with the current implementation of all Resident Intro members- no matter how many there are, nothing in the Constitution needs to be overridden.

venku122:
I recognize the issue, but I think we should keep in mind where these numbers came from. Freshmen should get a super majority of active members to sign their packet. This is reflected and rounded down in the 60% value. Freshman signatures don't matter at all.  It is my opinion and interpretation that gaining freshmen signatures is intended to compel freshmen to meet each other. I think we should not weigh a freshman's interaction between other freshmen during the packet process. Back in my day, I enjoyed meeting and socializing with the upperclassmen of CSH much more than my freshmen peers.

I think we should actively encourage freshmen to meet and interact with each other but have no official requirement for them to in our process. The opinions that matter of that of current members.

VictoriaWeaver:
@venku122  The basis of this amendment (along with the skewed percentages cause by freshmen signatures) is my disagreement with the statement that "Freshmen signatures don't matter at all."

A member would still have the right to ignore freshmen signatures at fall evals- this amendment does not take that away.  However, for the people that do see value in the freshmen signatures, they would still be able to use that information to cast their vote while also fixing the issue of skewing the overall percentage for an account.

mbillow:
For the same of keeping discussion on GitHub, I will mirror my comment here: @VictoriaWeaver I am fine with your amendment, except that we are including freshmen signatures. I would gladly close mine if yours didn't include freshmen in the process of account creation.

I still think they should be on packet, it is valuable if we stop telling people to just sign everyone's packet, but there is no reason a freshmen should be able to stop or help another freshmen in the process of getting their account.

TalCohen:
I completely agree with this @mbillow

VictoriaWeaver:
I don't understand why you value their signatures for one thing and not the other?  You say "there is no reason a freshmen should be able to stop or help another freshmen in the process of getting their account" but why wouldn't it be the same reason you count their signatures in the overall result?

mbillow:
@VictoriaWeaver Not at all. I view upperclassmen and alumni signatures (in the scope of account creation) as our experienced members agreeing that this person should have an account. Think of it as a good recommendation to the RTPs. At membership evaluations, I want to be able to look back at these numbers, but also see the larger picture of how they got along with their class, what events they attended, seminars, etc. In the same way RTPs don't care what social events you went to when creating your account, I don't think they should consider what freshmen did or did not sign your packet. Just because you did or didn't get the signatures of other intro members should not be factored into wether you get an account.

rswiernik:
@mbillow I actually think that your thought process is slightly flawed here. If you are using a metric, in this case freshmen signatures, to determine whether a person is member of this organization then you should very well use that same metric to determine whether that same person gets an account and access to house services. Sure the RTP's don't personally care, however that is not the point of this process. The RTP's simply facilitate the technical aspects of getting a freshmen to a point where they have full access to house resources. At the end of the day, that's what the packet/account creation process is doing - determining whether we want to continue with this persons evaluations process and give them full access to our services while we do it.

In the same vain, I dont think that these freshmen signatures should be ignored so early in the process and then used at evals, which comes significantly later in the process. This seems backwards to me, as at 10 weeks you, as a responsible social upperclassmen, should be able to see how this person interacts with the people on floor, whereas at 2 weeks in, you need to rely more on the faith of other people. I agree that we should not be telling freshmen just to sign other peoples packets, and I agree that they shouldn't be making the packet significantly harder, however just like an easy or stubborn upperclassmen, the evals director has the power to handle that situation.

gambogi:
tbqh I think this whole part of the intro process comes up a lot and I don't think it's really broken. 

- If you feel like the freshmen shouldn't be on the packet, continue not caring about missed freshman signatures at fall evals. 
- If you feel like the freshmen signatures should be more meaningful, continue to care about missed freshman signatures at fall evals.

Fall evals is somewhat subjective for a reason. I don't think asking for 60% is ever really that big a deal, and most upperclassmen are perfectly aware of the dynamics of the packet and are capable of voting accordingly.

liam-middlebrook:
@gambogi I believe that @mbillow's intent isn't focused so much on the Introductory Evaluation and packet completion in respect to it, but rather to the 60% required to get an account. If I recall correctly this year we have a large percentage of freshman in comparison to upperclassmen, such that every freshman signature would gain ~39-40% of packet completion. If I understand correctly one of (but not the sole intention) behind these amendments is to make the account creation barrier strictly require a certain percent of upperclassmen (be they on floor, or off/alumni/honorary), in order to prevent freshmen who shouldn't be trusted with access to certain house services / information from getting that access.

I believe this concept of restricted access is why @mbillow doesn't want freshman signatures to affect whether or not a freshman will get an account, this is because during the main packet season none of the on floor freshmen should have access to this information and do not understand the weight of their signature in that regard.

gambogi:
@liam-middlebrook please reserve spoonfeeding for children. I don't think asking for 60% is ever really that big a deal.

mbillow:
@gambogi, Liam's clarification on my point is correct. They are inherently different processes. You can fail packet and continue to Fall evals without an account, meaning the account has nothing to do with our evaluations process. We have just decided that the conclusion of packet is a good time for creation. With that said, I am still of the thought that freshmen can carry as much weight as each upperclassmen gives them in their own decision process at 10-weeks, but for account creation we should only be looking at their interactions with our full members.

gambogi:
I'm not sure why either of you think I didn't read the previous discussion
before commenting. I don't think asking for 60% is that unreasonable for
account creation. I'm not as bothered by the "power" freshmen have over one
another as some of you seem to be.

mbillow:
@gambogi If we are going to be asking for things that "don't seem unreasonable" then why not just tack on the requirement to attend two committee meetings and a social event. Just because a hurdle is low to the ground, doesn't mean it is in any way related. I am for only requiring the things that matter and leaving the rest of the requirements for Fall Evals.

VictoriaWeaver:
I hope you realize that "things that matter" is completely subjective.  This amendment makes sure that account creation isn't based on whether or not someone prioritizes one requirement over another.  That's why the distribution over each section is equal.

Fall evals on the other hand is subjective.  That's why we let upperclassmen use the data as they please to cast their vote.

gambogi:
Because at the end of the day this discussion just quibbling about a
fluctuating handful of upperclassmen signatures for an account. I don't
think it's critical that we have exactly X percent of upperclassmen sign
off on account creation. X +/- 5 is fine by me

stevenmirabito:
There seems to be two different discussions happening here and I think we've veered off the tracks a bit. From what I've read and heard, most of us agree that we're looking for some kind of change. While I don't really feel that the freshmen portion of packet produces any real value for anyone involved, compared to #75, it seems like this amendment is a decent compromise that both keeps the freshmen on the packet and emphasizes the importance of receiving a majority of upperclassmen signatures.

@mbillow This seems to achieve your goal of making upperclassmen signatures more valuable while appeasing those who feel the freshmen portion of packet is still valuable. Is the issue of having freshmen on the packet at all completely blocking you from reaching a compromise?

gambogi:
Failed at ballot.

PR #78, opened 2016-09-18
Revised Packet Composition

TalCohen:
Check one:

Summary of change(s):

Made the composition of the packet more specific so freshmen that are given Active Membership after their introductory evaluation are still considered in the same Introductory Members section.


VictoriaWeaver:
So this also officially allows Honorary and Advisory members to sign packets? Welp, that's awkward... +1 for fixing it now.

TalCohen:
Well, an interpretation of non-resident members could mean honorary and advisory, so I just wanted to be more specific about it here. This whole thing might be changed a little, and pushed in as a non-semantic change along with whichever amendment passes this week slightly changed a little

TalCohen:
Out of date.

PR #84, opened 2016-10-12
Change to qualifications of OpComm Director

bmbowdish:
Check one:

Summary of change(s):
I think that this is truly a non-semantic change, but I understand that we have to be cautious to the changing of the constitution so this will have to be amended the traditional way.  
This amendment changes the qualifications of the OpComm Director so that they are more consistent with the rest of the Executive Board's qualifications.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -297,7 +297,6 @@ They are generally very task oriented and are chaired by a House member.
 A directorship has some jurisdiction in its area of interest and often is responsible for the day-to-day decisions regarding its area of interest.
 Any large expenditures or large effect decisions must be brought before the entire House
 \asection{Members of the Executive Board}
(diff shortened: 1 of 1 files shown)


articles.tex
```diff
@@ -311,7 +310,8 @@
 \asubsection{Non-Voting Members}
 \begin{itemize}
 	\item Ad Hoc Directors
-\asubsection{House Secretary}
+        \item House Secretary
+        \item Chairman
```
liam-middlebrook: I would put Chairman, then House Secretary, then Ad Hoc Directors personally, but that's not required and can be changed non-semantically later.

liam-middlebrook:
Passed 29 - 0

PR #86, opened 2016-11-07
Amendment Proposal to requirements to get account

jmyates:
Check one:

Summary of change(s):

Amendment Proposal: If an introductory member fails to meet the requirement to get an account after they have completed the two week packet period, the introductory member may have the Evals Director bring up a vote at one preceding house member to give the introductory member an account. This vote may happen once during the introductory member's 10 week evaluation process. The vote is a simple majority barring meeting Quorum

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -362,13 +362,11 @@ Root Type Persons have the authority to manage user accounts for House systems. 
 \begin{enumerate}
 	\item Sign the Code of Conduct sheets pertaining to the responsible utilization of Computer Science House and Rochester Institute of Technology facilities.
 	\item Obtain greater than or equal to 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations.
-	\item An Introductory member who fails to meet the signature percentage requirement may be approved for an account by 50\% majority vote at a House Meeting.
-	This vote may be initiated by an Introductory member by notifying the Evaluations director of their intent after the completion of their packet.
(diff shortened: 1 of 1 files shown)


RamZallan:
> at one preceding house member

I think you mean "at one **following** house **meeting**," or something to that effect. Just pointing it out in case it causes any confusion for those who weren't there for your proposal at house meeting tonight.

jmyates:
The wording in the amendment itself says following. I copied in the wrong text to the comment identifying the change I believe.

VictoriaWeaver:
I have several problems with this.

With the current state of this amendment, there is no designated quorum, which allows the loophole of the Chairman calling an emergency house meeting with just him and the Evals Director, then either if them can grant the member an account.  

Adding a quorum requirement would fix the that loophole but it would also just make this more redundant with the 10-week vote.  We already grant accounts if a person passes their 10-week vote, which requires a simple majority with a quorum of 2/3 possible votes.

It's also easier by 10% (60% packet, 50% vote) to get an account in this manner than it is with packet.  If anything it should be more difficult (e.g. 80% vote).  

And currently, it's not essential for a member to have an account in the first 10 weeks, especially now that we have designated Slack channels for almost everything the intro member would need.

I also can't think of any cases in which I would vote yes for a member to get an account without passing either packet or 10-week.  Extenuating circumstances for performance on the packet should be taken up with the Evals Director, who can grant an extension if they see fit.

gambogi:
@VictoriaWeaver that's my bad. Noticed the quorum issue but forgot house meeting doesn't have an established quorum requirement (or formal definition). I chose 50% because it's one of the existing vote definitions. 

That said, totally on board with your reasoning. I'm not a fan of this proposal.

gambogi:
Ping if you want to reopen.

PR #88, opened 2016-11-07
New packet

bmbowdish:
Check one:

Summary of change(s):

Introductory members may redo packet before 10 weeks

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -178,8 +178,8 @@ Each Introductory Process participant, after the first week, is given two (2) we
 	\end{itemize}
 	\item A list of seven (7) annual House social events
 	\item A list of seven (7) major House technical achievements
-\end{itemize} \\
-At any point before the Introductory Evaluation, an Introductory Member may make a second attempt at the Introductory Packet. The results of this Introductory Packet may only affect the number of signatures evaluated at the Introductory Evaluation. 
+\end{itemize}
+At any point before the Introductory Evaluation, an Introductory Member may make a second attempt at the Introductory Packet by requesting a new Introductory Packet from the Evaluations Director. The results of the new Introductory Packet may only affect the number of signatures evaluated at the Introductory Evaluation and, as long as the Introductory Member receives the appropriate number of packet signatures that would normally be required in the \ref{Creation of Accounts} process. 
 \bsubsection{Expectations of an Introductory Member}
 Before the end of the Process, an Introductory Member is expected to:
 \begin{itemize}
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -178,7 +178,8 @@ Each Introductory Process participant, after the first week, is given two (2) we
 	\end{itemize}
 	\item A list of seven (7) annual House social events
 	\item A list of seven (7) major House technical achievements
-\end{itemize}
+\end{itemize} \\
+At any point before the Introductory Evaluation, an Introductory Member may make a second attempt at the Introductory Packet. The results of this Introductory Packet may only affect the number of signatures evaluated at the Introductory Evaluation. 
 \bsubsection{Expectations of an Introductory Member}
 Before the end of the Process, an Introductory Member is expected to:
 \begin{itemize}
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -179,7 +179,7 @@ Each Introductory Process participant, after the first week, is given two (2) we
 	\item A list of seven (7) annual House social events
 	\item A list of seven (7) major House technical achievements
 \end{itemize}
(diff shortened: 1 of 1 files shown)


liam-middlebrook:
I think that the introductory member should be able to receive an account if they get above a threshold on this (perhaps a slightly greater percentage would be in place?).

lueking:
Completing a secondary packet before ten weeks should have the same ramifications as completing the first one, seeing as that's the reason for this proposal.

VictoriaWeaver:
> may only affect

The "may" part is sketchy- I interpret this as "it can... but it might not".
What happens if the I do worse on my second packet- do I keep the better score or do I have to keep the second one?
Can I stop the second packet halfway through and keep my old score?

devinmatte:
I agree with the sentiment that victoria shares. If someone decides to redo their packet, and gets a worse score, this should replace their old packet score no matter what, as it is an optional packet.
It's a risk, but it's a risk that _could_ benefit them.
With that being said, if someone got 60% the first packet, but gets 50% the seconed time, should they then also lose their account since this score becomes their new score.
I think all of this should be written as part of the amendment for clarification.

ajgajg1134:
This is an interesting proposal, one of the main issues I identified with the process at least specifically this year was the lack of feedback to freshmen on their progress between the end of packet and ten weeks since it's a very difficult thing to gauge. This might be a good tool to help them gauge their own progress. 

I do have some concerns namely I think the packet is a tool and not a very good one. It works for the beginning of the year because a large number of freshmen are working together talking about it and all the upperclassmen know the packet is going on. Not to mention planned events around the packet like welcome back drawing a number of off-floors and alumni to floor to help facilitate those sometimes more difficult signatures. Floor at that point of the year is generally in a "get to know each other" mood as work loads haven't picked up.

I agree that we may have identified an issue with our process where people fail / do poorly on the packet, but I would much rather keep the current system of just dealing with it at 10 weeks through either a conditional to re-do the packet, passing them outright if they've gotten to know enough people over the 8 weeks afterwards, or failing them if their packet performance was not good at all in addition to other criteria generally. 

I worry that by allowing freshmen to more easily do the packet again we'll get a large number of people trying to do the packet a second time when in my mind we want to reduce the number of packets out. As has been identified the packet is stressful for everyone involved and is in no way perfect. Freshmen who have a lot of drive may think it's necessary with only say 8 missing signatures to try and do the packet again when they may clearly be on the edge of passing outright.

TL;DR this "issue" is mostly handled by just properly voting and discussing at 10 weeks. Less packets out = better.

nicholastmosher:
@devinmatte I don't think that the second packet should replace the first one, even if the outcome was worse. The point of this amendment is to give freshmen the opportunity to take initiative to improve their score (for the point of receiving an account or other reasons), but I don't think it would be beneficial to anybody to disregard the old score. I think that any and all packet attempts made should be recorded and presented at the 10 week evaluation for house to take into consideration.

To me, a reattempt of the packet would show great initiative on the freshman's part, especially since packet becomes increasingly difficult after the initial packet season. I don't think we should punish anyone for trying again.

VictoriaWeaver:
@bmbowdish while you now state that the second packet will affect the requirement at Intro Evals, you still fail to include how it will affect the requirement.  It is still unclear whether or not you intend to use the second packet to replace the original score completely, conditionally, or in some other manner.

bmbowdish:
I do not think that it is unclear. 

"At any point before the Introductory Evaluation, an Introductory Member may make a second attempt at the Introductory Packet by requesting a new Introductory Packet from the Evaluations Director. ... The results of the new Introductory Packet will also affect the number of signatures evaluated at the Introductory Evaluations."

This is pretty easy to understand for me. If you think it can be clearer, please give me the text that you feel is better.

devinmatte:
@bmbowdish I agree with @VictoriaWeaver That it is a little unclear. I think a better wording may be that:

> The results of the new Introductory Packet will replace the original score evaluated at the Introductory Evaluations.

Or 

> The results of the new Introductory Packet will be displayed alongside the original score when being evaluated at the Introductory Evaluations.

Or whatever the intended meaning of the amendment is. The language is a little vague.

VictoriaWeaver:
That's the point- I don't understand how the second packet affects the evaluation.  Is it replacing it no matter what?  Or is it only replacing it if it's a better score?

I would change this sentence
>The results of the new Introductory Packet will also affect the number of signatures evaluated at the Introductory Evaluations.

to something like this:
>The results of the new Introductory Packet will replace the results of the first attempt to be evaluated at the Introductory Evaluations [if the score is improved.  If the results of the did not improve, the original packet score is evaluated].

where the [] section depends on your intention.  Currently, the amendment does not specify how the packet is affect.   Simply stating that the results will be affected is not enough be cause the effect is not defined.  Maybe you want to average the two results instead of replacing them.

Is my complaint more clear?

liam-middlebrook:
Failed 4 - 34.

PR #90, opened 2016-11-10
Executive Board appeal to House

bmbowdish:
Check one:

Summary of change(s):
Creates appeals for Executive Board decision.

This is by Sam 👍

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -544,10 +544,14 @@ The vote of a vacated Executive Board position shall be cast as an abstention in
 \end{enumerate}
 
 \asection{Appeals}
+\asubsection{Executive Board Appeal}
 Decisions made by the Executive Board may be appealed and overturned.
 To initiate an appeal, a member must have the support of three voting members of the Executive Board, or a petition with the signatures of one-third of Active Members.
 After the appeal is presented, a Simple Majority vote is held by the Executive Board whether or not to overturn and reevaluate the decision.
 If the vote passes, the Executive Board may discuss and make a new ruling by another Simple Majority vote.
+\asubsection{House Appeal} 
+To initiate an appeal to House of an Executive Board decision, a petition with the signatures of two-thirds of Active Members must be brought to a House Meeting. After the appeal is presented and discussed, a Two-Thirds vote is held by the attending members. 
+If the vote passes, attending members may discuss and make a new ruling by a Two-Thirds vote. 
 
 % ARTICLE V - VOTING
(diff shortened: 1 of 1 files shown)


VictoriaWeaver:
Before I state my opinion on this, I think there's a better way to add this in since they both address appealing Eboard decisions.

I would suggest adding it under the Eboard appeals section:
> Decisions made by the Executive Board may be appealed and overturned.  The appeal may be initiated in one of two ways...

and then go on and list the two methods under the same subsection.

Concerning my opinion of this, I'm not a fan.  The fact that you need 2/3s of house to even consider an appeal, but then only need 2/3 of people attending to overturn the decision seems awkward- there's that whole "chairman calls an emergency house meeting" loophole.  Maybe add a quorum requirement? Though I'm still not sure how I'd feel about it.  

I'm also curious to know what inspired this so if that info is available please share.

gambogi:
Please add quorum requirement.

articles.tex
```diff
@@ -545,9 +545,14 @@
 
 \asection{Appeals}
 Decisions made by the Executive Board may be appealed and overturned.
+\asubsection{Executive Board Appeal}
 To initiate an appeal, a member must have the support of three voting members of the Executive Board, or a petition with the signatures of one-third of Active Members.
 After the appeal is presented, a Simple Majority vote is held by the Executive Board whether or not to overturn and reevaluate the decision.
 If the vote passes, the Executive Board may discuss and make a new ruling by another Simple Majority vote.
+\asubsection{House Appeal} 
+To initiate an appeal to House of an Executive Board decision, a petition with the signatures of two-thirds of Active Members must be brought to a House Meeting. 
+After the appeal is presented and discussed, a Two-Thirds vote is held by the attending members.
```
liam-middlebrook: Agreeing with @gambogi please add some kind of quorum here. Perhaps `A quorum of two-thirds of the Total Number of Possible Votes`

liam-middlebrook:
@bmbowdish are you considering updating this PR with the suggested changes? If you plan to do so let me know, other wise I plan to bring this up with the rest of Amendment Train #1

liam-middlebrook:
After discussing with a few people this amendment doesn't fit the spirit of an amendment train because it's not a fix based on an already common house practice. I've removed it from the train due to this.

gambogi:
@bmbowdish is this going to be brought up at some point?

bmbowdish:
Will ask Sam about it tonight

PR #92, opened 2016-11-14
Moves semantic change discussion from House Meeting to History Discussion

bgrawi:
Check one:

Summary of change(s):

Moves the amendment discussion from House Meeting to History Discussion instead. It's a more appropriate place to talk about amendments, does not bloat House Meeting unnecessarily, and still leaves the overall process in-tact.  

---

Interestingly, we have also mis-interpreted the constitution this whole time and discussion *is* supposed to take place when an amendment is proposed so that changes can be made ahead of time to the vote (instead of just voting no if you don't like the wording). This amendment would reinforce that this discussion take place at a History Discussion instead and that simply a summary of the discussion would be presented to House.

I still think discussion is key to making quality and useful amendments, I just feel that House Meeting is no longer the best place for them given the quantity and length of discussion needed the past two semesters. History Discussions are well suited specifically for talking about large issues that face House and will allow members who feel passionately about an issue to debate and present findings to House at large.

The timeline for an amendment would look like this:

1. A member creates a GitHub pull request for an amendment and notifies the History Director.
2. The amendment will be discussed at the next available History Discussion meeting.
3. The member can either take feedback from the meeting to make appropriate changes/leave as is before presenting to House, or they can decide to cancel the amendment altogether if discussion recommends that.
4. At the following house meeting, the amendment is posted to House along with a summary of the discussion leading up to the amendment. *Ballots are handed out now*.
5. Normal ballot voting proceeds, and then the amendment either passes or fails.

~~Make House Meetings Great Again~~

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -406,6 +406,7 @@ However, the Chairman and at least two-thirds of the Voting Members must be pres
 	\item To collaborate with the Social Director(s) to ensure that all the Active, Alumni, and Advisory Members are informed of upcoming events.
 	\item To oversee the production and distribution of a semi-annual newsletter.
 	\item To oversee the creation of a yearbook outlining House events for the year.
+	\item To preside over House History Discussion meetings, which must be held if there are upcoming \ref{Semantic Changes} to the Constitution but are otherwise held at the director's discretion.
 \end{enumerate}
 
 \asubsection{Responsibilities of the House Secretary}
--- a/articles.tex
+++ b/articles.tex
@@ -98,9 +98,9 @@ Alternatively, the change may be presented to the House for discussion followed 
 A quorum of fifty percent of the Total Number of Possible Votes is required for passage.
 
 \asubsection{Semantic Changes}
-Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
+Any semantic change to the Constitution requires the change to be proposed in writing for a History Disussion meeting.
 Any modifications made due to the discussion are added to the written proposal and the modified proposal is posted in the House during the week.
-The final proposal is presented the following week and ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
+The final proposal, along with a summary of the discussion, is presented the following week at House Meeting and ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
 The ballots are collected for a minimum of a forty-eight hour period.
 A quorum of two-thirds of the Total Number of Possible Votes must cast ballots for the vote to be official.
 A vote equaling or exceeding two-thirds of the number of votes cast is required for the change to be placed into the constitution.
--- a/articles.tex
+++ b/articles.tex
@@ -100,7 +100,7 @@ A quorum of fifty percent of the Total Number of Possible Votes is required for 
 \asubsection{Semantic Changes}
 Any semantic change to the Constitution requires the change to be proposed in writing for a History Disussion meeting.
 Any modifications made due to the discussion are added to the written proposal and the modified proposal is posted in the House during the week.
-The final proposal, along with a summary of the discussion, is presented the following week at House Meeting and ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
+The final proposal and a summary of the discussion are presented the following week at House Meeting and ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
 The ballots are collected for a minimum of a forty-eight hour period.
 A quorum of two-thirds of the Total Number of Possible Votes must cast ballots for the vote to be official.
 A vote equaling or exceeding two-thirds of the number of votes cast is required for the change to be placed into the constitution.


venku122:
I agree that having discussion the day of ballot release is not ideal. I believe there was some confusion last year regarding exactly when discussions were supposed to be held, and a few discussions had been accidentally skipped. The current system was a compromise out of that. The idea is to make the public text known and available for discussion and editing via github/webnews/discourse and ideally have a final text ready for house to formally discuss.

I disagree about discussion happening at History Discussion meetings. The current constitutional rules ensures that every constitutional rule change is brought before house and that every active member has a chance to discuss the amendment, and hear the pros and cons presented. Moving discussion to history meeting makes the discussion optional, reducing participating in the amendment process. 

People have various activities and responsibilities during the week, meaning not everyone can attend history discussion even if they wanted to. Opt-in discussions also allow for unbalanced discussions, where enthusiastic supporters or detractors can group together and attend the meeting, leading to discussions that don't represent the "average" opinion of house.

The bottom line is this amendment would reduce participation in the amendment process.

stevenmirabito:
I wholeheartedly disagree with this. Changes that affect all of house should be discussed with everyone present, during a block of time that everyone has set aside for that purpose. When this was tried briefly last year very few people attended the discussions and there was a lot of confusion regarding the amendments affected. I think this would be a serious detriment to our organization.

VictoriaWeaver:
History Discussion is not defined in the Constitution.  The History Director can choose to hold a discussion or not, but if the current director chooses not, then it does not exist.  There cannot be a requirement to hold discussion for amendments at a meeting that technically does not exist.

bgrawi:
Okay, clearly this needs more work, but I still think this is a huge issue facing house. [An amendment similar to this](https://github.com/ComputerScienceHouse/Constitution/pull/48/files) was brought up last semester and it failed due to similar wording issues. Those appear to have been fixed, but it looks like it was forgotten about. What are people's opinions about the current wording of that?

venku122:
@bgrawi The primary sentiment of these amendments in my opinion is to allow people to ignore the amendment process. CSH is a direct democracy, where voter participation is critical in every decision. 

If people are annoyed by the volume of amendment discussions, they have the ability to talk with the person who proposed the amendment up to a week before house meeting, discuss the amendment, and make a case for not bringing the amendment up to vote at all. 

Constitutionally ratifying voter apathy is not a good idea.

gambogi:
@VictoriaWeaver if this was phrased in such a way that the History director was required to hold a discussion of the amendment the definitional issue can be avoided

VictoriaWeaver:
@gambogi if that were to happen, I would rather this wait to be enacted for the next academic year so that those running for History are aware of all the responsibilities of the position.

bgrawi:
@venku122 I disagree that the sentiment is to allow people to ignore the process. I intended the sentiment to be around making House Meetings better value a member's time. The current situation is not ideal as the constitution has never gone under this much change before. I think we should at least do *something* to address the problem. I view it as a real and existential threat to CSH honestly. If House Meetings are not worth people's time, then literally the only option is to not attend House Meeting, thus failing membership requirements and disenfranchising members who would otherwise be fine with staying an active member.

This also still has the same voting requirements with regards to quorum as before, so this has little to do with voter apathy and more to do with the fact that a majority of house will have an opinion on a topic that is unlikely to be swayed, even after 45 minutes of discussion. Would adding a constitutionally-defined time limit help? I don't think so because you will wind up stopping potentially valid points from coming up. I also don't think relying on the chairman to "be good at moderating" is a solution either because then it is wholly dependent on the year/person if meetings are long or not.

I would also like to clarify that I am not against the volume of amendments that came up this year (many are very positive!), just the time spent at House Meeting literally every week that can last up to an extra hour. We only recently, like 2 years ago, added that extra hour of 1250 reservation because of 10 week votes or elections which had to overflow into the lounge, *not because of amendments*. House meeting should **not** routinely take 2 hours. I am very worried that CSH is loosing its focus and is becoming primarily a self-serving political system over all other qualities.

@VictoriaWeaver Regardless, I've added a first draft of how I think History Discussions should be defined in the constitution to fix this. I think changing requirements during an academic year is not that large of an issue, but if the history director at the time this is enacted has any input on this regard, then they are free to voice it. As it stands though, the only functional difference from the way current history discussion meetings are held is that the meeting is mandatory when there is an amendment. Since these meetings generally happen every week anyways there are no actual difference in time commitments.

articles.tex
```diff
@@ -98,9 +98,9 @@
 A quorum of fifty percent of the Total Number of Possible Votes is required for passage.
 
 \asubsection{Semantic Changes}
-Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
+Any semantic change to the Constitution requires the change to be proposed in writing for a History Disussion meeting.
```
bgrawi: This is actually different than my original intentions of it being proposed to just the History Discussion meeting itself, but I can see concerns with people not even knowing about it to attend the meeting. In my head I was imagining it working like how current discussion topics are announced. I'll have to figure out the best way to word this though, probably say:
```
...requires the change to be proposed in writing at a House Meeting.
The amendment will be discussed at the following History Discussion meeting.
```
or something like that depending on what people like.

gambogi:
@bgrawi do you plan on brining this to vote? If not I'm going to close.

PR #93, opened 2016-11-14
Revoke membership of advisors who no longer work for RIT

stevenmirabito:
Check one:

Summary of change(s):

Currently, once an Advisor gains Advisory Membership, they keep it indefinitely, unless they explicitly resign. This amendment modifies the Advisory Membership Term section of the Articles to define that if their employment by RIT is terminated _or_ they resign, their membership should be revoked.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -268,7 +268,7 @@ Advisors are not subject to formal evaluations.
 An Advisor may resign by submitting in writing the reason for resignation to the Chairman.
 The resignation will be read at the following House Meeting and become effective at that time.
 \asubsection{Advisory Membership Term}
-Advisory Membership shall last until the member resigns.
+Advisory Membership shall last until the Advisor no longer meets the qualifications as defined in \ref{Advisory Membership Qualifications} or until the Advisor resigns.
 
 \asection{Leave of Absence}
 A leave of absence offers the option for members to take extended time away from their responsibilities to House for any number of personal issues including, but not limited to, physical illness, mental illness, or care giving for a sick family member. Computer Science House recognizes the need for its members to take care of their personal well-being and will support them through the process of requesting a leave of absence as well as acclimating back to the culture of House upon their return. Upon approval of a leave of absence request, the Evaluations Director is notified and the member's leave is applied immediately. A member is also allowed to be physically present and also be on a leave of absence. 


VictoriaWeaver:
This might just be me, but the current wording kind of makes it sound like the advisor has to be fired/laid-off by RIT.  I would rephrase it to 
>until the member resigns or no longer satisfies the qualifications of the role etc.  

Other than that it just seems a little frivolous to bring this up, so I'd like to hear reasoning for doing so.

stevenmirabito:
@VictoriaWeaver That's a good idea. I've updated the wording to reference the qualifications section and added "current" to that definition to make it a bit more explicit.

This is being brought up because we realized today that we have a bunch of advisors who still have active CSH accounts who don't work for RIT anymore, and no constitutional way to revoke their membership without them explicitly resigning.

articles.tex
```diff
@@ -253,7 +253,7 @@
 
 \asection{Advisory Membership}
 \asubsection{Advisory Membership Qualifications}
-Advisory Membership is open to all members of the Rochester Institute of Technology professional, academic, or administrative staff.
+Advisory Membership is open to all current members of the Rochester Institute of Technology professional, academic, or administrative staff.
```
gambogi: I don't think current is necessary here. You either are a member of the staff or you aren't.

articles.tex
```diff
@@ -268,7 +268,7 @@
 An Advisor may resign by submitting in writing the reason for resignation to the Chairman.
 The resignation will be read at the following House Meeting and become effective at that time.
 \asubsection{Advisory Membership Term}
-Advisory Membership shall last until the member resigns.
+Advisory Membership shall last until the Advisor no longer meets the qualifications as defined in \ref{Advisory Membership Qualifications} or until the Advisor resigns as defined in \ref{Advisory Membership Resignations}, at which time their membership is revoked.
```
gambogi: I think you can skip the `as defined in .... revoked.` 
The current wording could be construed to mean that when and advisor resigns or leaves RIT they also lose their existing "normal" membership.

articles.tex
liam-middlebrook: :+1: Could be confusing in combination with sabbatical.

liam-middlebrook:
@stevenmirabito are you considering updating this PR with the suggested changes? If you plan to do so let me know, other wise I plan to bring this up with the rest of Amendment Train #1

stevenmirabito:
@liam-middlebrook @gambogi Updated this PR with the suggested changes and synced with upstream.

devinmatte:
This amendment obviously brings up a lot of discussion. As was outlined today I think this amendment needs to establish some form of membership or access for previous advisers.
The main problem brought up in discussion was the lack of communication with past advisers.

I am not sure what the proper status for past advisers is, however in order for I to vote yes on this amendment there would need to be a status of membership after resignation from RIT or advisory positions. They should be able to continue to communicate with us, and continue to be resource for us, unless we explicitly do not want them to.

nickdepinet:
Is there any plan to update this PR with the changes from the discussion at house meeting before it comes up to ballot?

stevenmirabito:
Based on the discussion that we've had during house meeting, here, and on Discourse, I will not be bringing this amendment to ballot. Thank you to everyone for your input.

PR #94, opened 2016-11-14
Semantic change discussion clarification

sgreene570:
Check one:

Summary of change(s):

Exactly when house wide discussions of semantic changes should be held is currently vague and unclear.  Currently, we have discussion on amendments one week after their proposal.  The current wording could easily be interpreted as meaning otherwise.  This change would require proposed amendments to be discussed immediately after being proposed, rather than one week later.  This change would prevent members from voting no on an amendment simply because they don't like the wording of the amendment as they would have been able to discuss their issues involving wording and phrasing one week prior to voting.  
As is, the constitution says that after an amendment's discussion, the amendment creator will have one week to edit and re work the amendment.  Currently, there is absolutely no time for a change to be made to amendment following it's house wide discussion.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -98,7 +98,8 @@ Alternatively, the change may be presented to the House for discussion followed 
 A quorum of fifty percent of the Total Number of Possible Votes is required for passage.
 
 \asubsection{Semantic Changes}
-Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
+Any semantic change to the Constitution requires the change to be proposed in writing at a House Meeting.  
+Discussion of the wording and the implications of the change should be held immediately following the proposal.
 Any modifications made due to the discussion are added to the written proposal and the modified proposal is posted in the House during the week.
 The final proposal is presented the following week and ballots are distributed for a ballot House vote as described in \ref{Balloted Vote}.
 The ballots are collected for a minimum of a forty-eight hour period.


articles.tex
```diff
@@ -98,7 +98,8 @@
 A quorum of fifty percent of the Total Number of Possible Votes is required for passage.
 
 \asubsection{Semantic Changes}
-Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
+Any semantic change to the Constitution requires the change to be proposed in writing at a House Meeting.  
+Discussion of the amendment's wording and implications will be held imeadiately following the amendment's proposal.
```
stevenmirabito: "imeadiately" -> "immediately"
"will be held" -> "should be held"

I would also change the wording of this to "Discussion of the wording and implications of the change [...]" as we don't use the word "amendment" here.

VictoriaWeaver:
Correct me if I'm wrong, but I think that the goal of having discussion after the edits were made is to ensure that people know what exactly they're voting on.  This prevents people from proposing an amendment, then changing it drastically after the initial discussion, and proposing an entirely different amendment to House without explicitly telling anyone.  Unfortunately it's not safe to assume that people will read their ballots before submitting them, especially if they think they already know what it's about.

I see the merits of having discussion both directly after proposal and before submission of ballots (and House probably will not like having both, so that's not really an option).  In the current state of affairs, I think that the most important edits (e.g. closing loopholes and clarifying intentions) already occur without immediate discussion (on github, on slack, etc).  

I personally prefer having House discuss the final form of the amendment so that the idea rather than the semantics are debated.  If the creator of the amendment feels strongly enough that changes should be made following House discussion, they can pull the ballot.

gambogi:
This looks like exactly the same process that's already described in the section, and therefore questionable as a semantic change. You guys just seem to not be following what's currently on the books.

sgreene570:
From what I have been told, the reason we aren't following the discussion policy set in the constitution is because the Chairman has chosen to interpret the semantic change section differently, potentially as a result of it be very vague.  This proposed change will remove any doubt of what the constitution intends for.  If this amendment were to pass, it would alter how house meeting is run, which is why I decided to mark it as a semantic change.

gambogi:
Hopefully not necessary in the future

PR #99, opened 2017-04-28
Changed Introductory Project Evaluations Process

sgreene570:
Check one:

Summary of change(s):
Modified introductory project section to grant the Evaluations Director control in deciding and reporting introductory project results for each individual introductory member.

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -146,7 +146,7 @@ The Executive Board may hold a closed ballot Simple Majority vote to extend the 
 \bsubsection{The Introductory Project}
 Introductory members are expected to demonstrate initiative and an ability to work well in a group environment through the Introductory Project.
 \bsubsection{Introductory Executive Board}
-An Introductory Executive Board is elected by the introductory members to lead and manage the project, represent the introductory project participants to House, and evaluate participants based on their contribution to the project.
+An Introductory Executive Board is elected by the introductory members to lead and manage the project as well as represent the introductory project participants to House.
 The Introductory Executive Board consists of three members:
 \begin{itemize}
 	\item A President, who leads the project and acts as primary representative to House through weekly status reports at House Meeting.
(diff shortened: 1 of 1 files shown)


devinmatte:
As much as I think having the Evals director present the results of the intro project (Freshman Project) makes sense. I think it makes more sense for the Froshboard to continue to present the results. Although it is a stressful experience presenting the results to their fellow intro members, it's also a valuable one. It's also a task that isn't necessary for the Evals director to do, especially since they already present to each freshman the result of their 10 week evaluations. That would require the evals director to meet with each freshman twice in one week. I think changing that portion of this amendment makes it more appropriate.

bylaws.tex
```diff
@@ -161,6 +161,9 @@
 \bsubsubsection{Introductory Project Advisors} 
 The Evaluations Director may choose between one (1) and three (3) House members to be Introductory Project Advisors.
 The advisors assist in organizing the participants at the beginning of the Introductory Project and thereafter act as a resource to keep the Project running smoothly.
+\bsubsubsection{Introductory Project Evaluations Process}
+The Evaluations Director determines whether each introductory member passes or fails the Introductory Project by using information gathered from the Introductory Project Executive Board.
```
devinmatte: I think it's more valuable for the Freshman Executive Board to determine who passes/fails instead of the Evaluations Director. Freshman elected these representatives, house elected the Evaluations director, however they also get a vote on the intro members. The Executive board represents the Freshman's collective voting power.

gambogi: I would add some mention which includes project advisor input in addition to the freshman eboard.

bmbowdish: Giving 4 freshmen an opportunity to show that their grown ups isn't worth the risk of them being completely incapable of telling their peers about their failures.

bmbowdish: They're

gambogi: You were right the first time

bmbowdish: Seems good 👍

PR #102, opened 2017-05-04
collected -> cast

bmbowdish:
Check one:
- [ x] Non-semantic Change: Spelling, grammar, or formatting changes.

Summary of change(s):

Fixed small thing brought up by @nick


jmyates:
This change is not a non- semantic change. Per the constitution a cast vote does not include an abstention vote even if that vote is cast/ inserted into the ballet box. Therefore blank ballets or abstention ballets placed in the ballet box would not count towards this which was the whole point of the judicial.

bmbowdish:
An Abstention is defined as a vote indicating a neutral position in the vote.
A means to abstain must always be provided in a vote.
Abstentions are counted towards a Quorum, but not towards the Total Number of Votes Cast used to determine if a vote passes or not.

The judicial determined that there is no way to abstain in the alternative voting method...
There are issues, but this amendment doesn't change whats already in the constitution right now.

PR #105, opened 2017-06-20
Corrected Non-Voting Members section and removed trailing whitespaces

sgreene570:
Check one:

Summary of change(s):
Changed Non-Voting Members section of the Executive Board to reflect the fact that there can be more than one ad hoc director at a time.
Also removed ALL trailing whitespaces from .tex files.


liam-middlebrook:
Can you split these into two separate commits and PRs?

PR #106, opened 2017-06-20
Corrected singular noun used in non voting eboard member section

sgreene570:
Check one:

Summary of change(s):
Fixed non-voting member list under the executive board section to reflect the fact that more than one ad hoc director can exist at once.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -324,7 +324,7 @@ Any large expenditures or large effect decisions must be brought before the enti
 \begin{itemize}
 	\item Chairman 
 	\item House Secretary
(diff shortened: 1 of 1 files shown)


articles.tex
```diff
@@ -324,7 +324,7 @@
 \begin{itemize}
 	\item Chairman 
 	\item House Secretary
-	\item Ad Hoc Director
+	\item Ad Hoc Director (s)
```
liam-middlebrook: Shouldn't it be `Director(s)`?

PR #113, opened 2017-09-05
ran script to add periods to end of any bullet points

SethGower:
Check one:

Summary of change(s):
Ran a python script to add a period to the end of any bullet points or enumerated items that don't have a period at the end.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -457,9 +457,9 @@ However, the Chairman and at least two-thirds of the Voting Members must be pres
 \end{enumerate}
 \asubsection{Qualifications to be the Operational Communications Director}
 \begin{enumerate}
-	\item Candidates must be a current Root Type Person \ref{Operations of the Operational Communications Directorship}
-	\item Candidates must be a current Active Member
-	\item A Root Type Person cannot be the director if they currently hold any other voting Executive Board Position
+	\item Candidates must be a current Root Type Person \ref{Operations of the Operational Communications Directorship}.
+	\item Candidates must be a current Active Member.
+	\item A Root Type Person cannot be the director if they currently hold any other voting Executive Board Position.
 \end{enumerate}
 
 \asubsection{Qualifications to be the House Secretary or an Ad Hoc Director}
--- a/articles.tex
+++ b/articles.tex
@@ -1,8 +1,8 @@
 % BEFORE CHANGES ARE MADE TO THIS DOCUMENT:
-% -References will be automatically updated if any part is added, deleted, etc. 
-%  However, if a sub part is moved to a different part, its references must be 
+% -References will be automatically updated if any part is added, deleted, etc.
+%  However, if a sub part is moved to a different part, its references must be
 %  changed.
-% -This document must be ratified by the House (as per the Constitution), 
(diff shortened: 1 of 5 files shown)


bylaws.tex
```diff
@@ -164,32 +164,32 @@
 \bsubsection{The Introductory Packet}
 Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain the following:
 \begin{itemize}
-	\item A signature from all Active/Introductory Resident members, each Executive Board member, and fifteen (15) of any combination of the following:
+	\item A signature from all Active/Introductory Resident members, each Executive Board member, and fifteen (15) of any combination of the following:.
```
liam-middlebrook: Already has punctuation.

devinmatte: There are a few issues like this throughout the document. Might want to modify the script a bit more.

articles.tex
```diff
@@ -312,19 +312,19 @@
...
 \asubsection{Voting Members}
 \begin{itemize}
-	\item Evaluations Director
-	\item Social Director(s)
-	\item Financial Director
-	\item Research and Development Director(s)
-	\item House Improvements Director
-	\item Operational Communications Director
-	\item House History Director
+	\item Evaluations Director.
```
liam-middlebrook: I'm not sure that these need `.`'s

bylaws.tex
```diff
@@ -164,32 +164,32 @@
...
-		\item Alumni members (not including Non-Resident Executive Board members)
-		\item Honorary members
-		\item Advisory members
+		\item Non-Resident members who have passed a Membership Evaluation.
+		\item Alumni members (not including Non-Resident Executive Board members).
+		\item Honorary members.
+		\item Advisory members.
 	\end{itemize}
-	\item A description of each Executive Board position, including:
+	\item A description of each Executive Board position, including:.
```
devinmatte: This is another example of an issue with the script

SethGower:
Ok, so I will make it check for other kinds of punctuation, and should I have it check for how many words are in the `\item`?

bylaws.tex
SethGower: ok, so remove all periods?

devinmatte: I think that's up to discussion. I personally don't see this as an issue with the constituion. But at the very least we shouldn't have periods on the same lines as `:`

SethGower: Ok, I am doing this because my options were this or to fix a semantics problem that has been up since april. I didn't feel comfortable doing that, and figured that this was a low effect type of situation

periods.py
```diff
@@ -0,0 +1,17 @@
...
+        new_articles.write(line)
+
+original_bylaws = open('bylaws.tex', 'r')
+new_bylaws = open('tempBylaws.tex', 'w')
+for line in original_bylaws:
+    if('\\item' in line and line[-2] == '.'):
+        line = line[:-2]
+        new_bylaws.write(line + '\n')
+    else:
+        new_bylaws.write(line)
```
devinmatte: We do not need to version control this script. Remove this file.

gambogi:
wild: you can resolve conflicts from the web now

PR #115, opened 2017-09-24
Addition of rules for the handling of bullying, abuse, and judicial proceedings

metal-crow:
Check one:

Summary of change(s):
Improve handling of abuse and bullying claims, how the Judicial occurs for them, and the outcome of the judicial. This is done to prevent the silencing of abuse, and sidelining of bullying issues. Also serves to ensure impartiality and prevent the dropping of cases due to abusers in positions of power.

Change handling of cases where a Judicial occurs involving an e-board member. Needed because it is unlikely e-board can remain impartial when judging another e-board member.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -668,13 +668,18 @@ In the event of a tie in an Executive Board vote, the Chairman may cast the tie-
 % ARTICLE VI - JUDICIAL
 \article{Judicial}
 The Executive Board may approve (by simple majority) a request (from a member) for a Judicial proceeding when an official clarification of the Constitution or By-Laws is required, there is a conflict among House Members, or a House interest needs resolution.
+If a member comes forward with allegations that abusive speech, bullying, or threats are occurring to them from another member, a Judicial proceeding is immediately started, without the need for Executive Board approval.
+In the case of a member coming forward with an accusation of assault or more serious actions, the matter and all information known about it is immediately forwarded to The Department of Residence Life and/or the relevant department at RIT, bypassing any Executive Board decisions.
 \asection{Formation of a Judicial Board}
 A Judicial Board is made up of the Chairman, the Evaluations director, and an additional voting member of the Executive Board chosen by a majority vote among the Executive Board.
-If either the Chairman or Evaluations Director are deemed biased or unfit for a position on the Judicial Board, they will be replaced by another voting member of the Executive Board by means of a simple majority vote amongst the remaining Executive Board members.
+If either of the two non-Chairman member of the Executive Board are deemed biased, unfit for a position on the Judicial Board, or are involved in the Judicial as an accuser or accused, the Judicial Board will instead be composed of 2 non-Executive Board members randomly chosen from active on-floor members, and the Chairman.
+If the Chairman is deemed biased, unfit for a position on the Judicial Board, or involved in the Judicial as an accuser or accused, the Judicial Board will be composed of 3 non-Executive Board members randomly chosen from active on-floor members.
 \asection{Judicial Investigation}
 The Judicial Board will be responsible for making all necessary inquiries into the matter of the request brought to the Judicial Board.
+For Judicial proceedings initiated from a claim of abusive speech, bullying, or threats, the accuser is able to directly introduce evidence (including witnesses) to the meeting the Judicial decision occurs at. It is also required that all witnesses be interviewed by the board during the decision making meeting.
 \asection{Judicial Ruling}
 The Judicial Board may present an official ruling following the Judicial Investigation.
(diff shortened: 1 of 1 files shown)


stevenmirabito:
My immediate reaction is that we do not want to touch this. Anything this would apply to should be immediately taken to ResLife.

sgreene570:
I am on the same page as @stevenmirabito . Hard pass.  We should tread carefully.

metal-crow:
I see. I agree, but that should be codified then, because as is that is not guaranteed to happen. I can rework this to specify that

jehling:
**In response to line 675:** 
_675: +If either of the two non-Chairman member of the Executive Board are deemed biased, unfit for a position on the Judicial Board, or are involved in the Judicial as an accuser or accused, the Judicial Board will instead be composed of 2 non-Executive Board members randomly chosen from active on-floor members, and the Chairman._

In the event that an e-board member is accused, I do not believe selecting a random member of house to fill their slot is the best course of action. Perhaps instantiating a list of approved arbitrators ("approved" meaning voted on by house as being worthy of the position) would help to guard against this scenario by replacing the defendant with an approved, impartial third party.

metal-crow:
That is an excellent idea, thank you! I didn't like the random selection, but hadn't come up with a better idea.

mbillow:
Regardless of my opinions on this: `:s/Department of/Center for/g`.

brandonhudson:
I am incredibly opposed to this for several reasons:
- Given our continued dialogue with residence life over the past few years, we are **heavily** moving away from using our Judicial process to handle such events, and are instead immediately directing the situation to residence life. This is not our eboard's decision on the matter, but rather guidance coming directly from our reslife advisor and other members of the RIT staff.
  - I am personally of the opinion that once residence life concludes their work and an outcome is reached, we can then look at what they discovered and work hand-in-hand with our advisor to take care of internal actions that might be needed.
- I am **extremely** opposed to codifying any form of immediate membership removal that relies solely on one individual's testimony of discomfort. Removal of membership should be used with extreme care and should **never** be written as a pseudo-automated process in our constitution.
  - The cases you are citing as extreme here should never be acted upon in a "Judicial-first" manner - all of these should be immediately sent to residence life without any of our intervention.

mbillow:
Now for my opinions... Eboard last year spent an incredibly long time talking about this and how we should be using Judicial Boards going forward. We made sure to document that conversation for future generations, but basically the gist of it was: Any issue that involves illegal activity or harassment is not an issue CSH should be involved in resolving. The Center for Residence Life (CRL) in collaboration with the Center for Conduct and Conflict Resolution have a far more appropriate skill set in dealing with these issues. There are educations programs that people can be referred to, counselors that can be made accessible, etc. - none of which CSH has the resources to provide (nor should we). 

Codifying this though, can be hard. The purpose for the constitution is to be a guiding document that we refer to in our decision making processes, not a playbook with all of the answers. Harassment and assault are not transgressions that should be taken lightly and if they are handled incorrectly can mature into even larger issues. This document should be about operational guidelines and with that being said, if you want to codify the inappropriate nature of using a Judicial Board in these cases, we can talk about that further. As for handling these issues internally, I am unequivocally apposed.

speakerbug:
Just my two-cents about this whole amendment. It's possible this is completely irrelevant as I don't know the current state of floor, but thought I'd still post anyway.

I agree with @stevenmirabito that putting something like this in writing is putting too much responsibility for dealing with these kinds of things onto CSH. I do, however, understand the want for behavior like this to not be tolerated and handled appropriately. I'm not really sure of the current issues/temperament of floor, but I'm assuming this was proposed due to some sort of incident.

If this is the case and you feel like it's being incorrectly handled, you should reach out to someone in ResLife. I also think that instead of an amendment, a sort of "cheat sheat" could be created for dealing with issues like these with a partnership between ResLife and CSH. It would be super simple to create a document that has a list of all the people within RIT you can contact about certain issues that arise (like bullying/abuse, but also not just limited to that) and make sure everyone knows those resources exist. Announce it at house meeting and put the document up around floor. This will probably also be more beneficial in the end too because let's be honest, only a handful of people actually read this constitution, but everyone would read a quick "cheat sheet".

These resources are available at RIT because college is hard and bad things happen and some issues are just beyond the scope of EBoard. For me, bullying/abuse lands in that category.

TL;DR: Don't make this an amendment and instead create a "cheat sheet" with all the resources on campus to handle these types of issues and spread it around floor.

metal-crow:
I agree with much of the criticism posted here, and as such will be dropping the amendment as is. I may rewrite it in the future to enforce what seems to be the majority opinion (all cases such as these should be immediately sent to res life), but it will be dramatically rewritten.

PR #116, opened 2017-09-24
By-Law VI: Use of Intellectual Property

stevenmirabito:
Check one:

Summary of change(s):

Adds an intellectual property policy to the By-Laws.


BionicPi:
The way this was discussed in house meeting suggested that the issue was the publishment of material, at one point referenced to specifically as 'public use' rather than 'use' as is written in the amendment. Could the two (discussion of the amendment and the written amendment) be more closely aligned, as this is a not insignificant change in meaning.

gambogi:
- `prior` is unnecessary. 
- "Written permission" is not a defined concept in this document. You should specify that an eboard vote must be taken, or otherwise disambiguate the approval process as you see fit.
- You should consider consulting RIT/a lawyer if you want this to be useful for anything other than intimidating members.

liam-middlebrook:
Yes: 18 No: 15
FAILED

PR #117, opened 2017-09-28
Updated Expectations of House Members

zthart:
Check one:

Summary of change(s):
Update Expectations of House Members to reflect current practice

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -302,7 +302,6 @@ If the member is still unsatisfied after being heard by the Executive Board, the
 
 \bsection{Expectations of House Members}
 Each member is required to pay House dues as stated in \ref{Collection of House Dues}, attend all House Meetings, and attend at least twenty-five (25) of the permanent director meetings.
-Members are also required to actively participate in a majority of House activities.
 \\* \\*
 The member must participate significantly in at least one major House project during the academic year.
 The member is required to submit a description for this major project to the Executive Board for approval.


bylaws.tex
```diff
@@ -301,10 +301,9 @@
 If the member is still unsatisfied after being heard by the Executive Board, the appeal may be brought to the attention of the ResLife Advisor.
 
 \bsection{Expectations of House Members}
-Each member is required to pay House dues as stated in \ref{Collection of House Dues}, attend all House Meetings, and attend at least twenty-five (25) of the permanent director meetings.
-Members are also required to actively participate in a majority of House activities.
+Each member is required to pay House dues as stated in \ref{Collection of House Dues}, attend all House Meetings, and attend at least twenty-five (25) meetings consiting of the seven permanent director meetings and open Executive Board as defined in \ref{Open Executive Board}.
 \\* \\*
-The member must participate significantly in at least one major House project during the academic year.
```
zthart: I feel like this change alone could be made as non-semantic, considering this wording doesn't appear to show itself anywhere else in the document and is inconsistent with the following line. If the rest of these changes are shat upon I'll probably just make a PR for that.

liam-middlebrook:
Can you rebase+squash these into a single commit?

zthart:
I was hoping for that dank squash+merge but I guess we can't always get what we want

articles.tex
```diff
@@ -326,6 +326,10 @@
 	\item House Secretary
 	\item Ad Hoc Director(s)
 \end{itemize}
+\asection{Open Executive Board}
+Open Executive Board Meetings are open to all members of the House, and those with the express permission of the Executive Board.
+The goals and pupose of the open Executive Board meeting should be established by the Executive Board at the beginning of the standard operating session as defined in \ref{Standard Operating Session}, and relayed to all members of the House at the following House Meeting.
```
csssuf: s/pupose/purpose/

zthart: Spelling is difficult

devinmatte:
My two cents on this amendment are that I don't think Open Eboard should have attendance attached to it. My reasoning being that Open Eboard is often used to summarize the events of each meeting (which we also do in house meeting) and then allows for some discussion within eboard. Now the difference between open eboard, and a normal directorship meeting is that the participation from other members in the meeting is minimal. If I go to History, then there is a focus on discussion, however open eboard is very much different from that as the focus is on eboard members, and not wide discussion (generally).
I dont think giving attendance is necessary

bylaws.tex
devinmatte: If you could separate out your non-semantic changes into a separate PR, we could at least get those changes merged in while we wait to vote on the rest.

zthart:
My motivation for this is that there is no difference between the member that sits quietly (without participating in discussion) through history or financial or imps or any other meeting and members who sit quietly through open eboard - beyond that, there are times where some directorship meetings last only as long as attendance takes, whereas eboard always has some content.

To make the other argument, If someone can only attend committee meetings on thursdays (or whatever day eboard happens to be on) they have half the opportunities to get attendance than those who might only be able to attend meetings any other single day of the week.

zthart:
Granted - the days of the week argument is only effective when the committee meeting calendar looks the way it does now (although the past three years there has been relatively little movement).

zthart:
I've done the needful and moved the non-semantic major project wording change to a different PR - I'll make some changes to this one later to reflect that

zthart:
Alright, I figured I'd just make these changes piecewise because they don't really have that much in common beyond taking place in the same section of the constitution.

This now simply removes a line that states that members are expected to participate in a majority of house activities - this requirement is entirely subjective and the objective metrics of committee meeting attendance, house meeting attendance, and the major project are suitable data points for "participation", in my mind.

PR #121, opened 2017-09-28
bylaws: Update standard operating session length

csssuf:
Check one:

Summary of change(s):
RIT now uses a 14-week semester schedule. Update the length of the standard operating session accordingly.


gambogi:
ok so you're correct this is semantically different but I would feel really bad if everyone actually voted on this. Maybe we should broaden the authority of maintainers to include factual inconsistencies? 

In any case if you fix whatever conflicts you have I'll merge this.

csssuf:
I don't think there are any conflicts?

PR #122, opened 2017-10-02
Clarify membership meeting requirements

sgreene570:
Allow ad-hoc directorship meetings to count towards both
the introductory evaluations process as well as the spring
evaluations process

Wording can change need be

Check one:
- [X] Semantic Change: something about the meaning of the text is different
- [ ] Non-semantic Change: Spelling, grammar, or formatting changes.

Summary of change(s):

PR #124, opened 2017-10-30
Scale directorship meeting requirement by semester

csssuf:
Check one:

Summary of change(s):
Scale the permanent directorship meeting requirement based on how many semesters a member is active for. This moves the requirement in line (more or less) with the requirement for introductory members (one per week).

Things to note:
* I made the requirement fifteen per semester rather than one per week for a few reasons. The primary reason is that scaling it per week seems too fine-grained to me, and requires considerable extra bookkeeping. Scaling by week also makes the requirements per semester 14, which isn't quite as round a number as I'd like. It also made the wording slightly more cumbersome.
* I'm not convinced the wording of my change is ideal, so I'd welcome feedback on how to improve that. In particular, I want to make sure it doesn't read as requiring meeting attendance to actually be distributed to 15 per semester - just a total of 30 for the whole year, if you're active both semesters.

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -307,7 +307,7 @@ If a member disagrees with the outcome of any evaluation, (e.g. is not asked to 
 If the member is still unsatisfied after being heard by the Executive Board, the appeal may be brought to the attention of the ResLife Advisor.
 
 \bsection{Expectations of House Members}
-Each member is required to pay House dues as stated in \ref{Collection of House Dues}, attend all House Meetings, and attend at least twenty-five (25) of the House directorship meetings (including permanent and Ad-Hoc directorship meetings).
+Each member is required to pay House dues as stated in \ref{Collection of House Dues}, attend all House Meetings, and attend at least fifteen (15) of the House directorship meetings (including permanent and Ad-Hoc directorship meetings) for each Academic Semester in which they are an Active Member.
 \\* \\*
 The member must participate significantly in at least one major project during the academic year.
 The member is required to submit a description for this major project to the Executive Board for approval.


csssuf:
Hmm, the changes from your PR seem to have mysteriously disappeared. I suspect #117 was merged after #122 and accidentally ate those changes. This will need to be resolved separately. My amendment isn't intending to change that, in any case. I do think the wording _is_ clear on requiring 30 meetings for those who are present for fall and spring - what makes you think it isn't?

csssuf:
Wow, Github's web UI for resolving conflicts is hot trash. Gonna force-push a real fix.

zthart:
I don't know if this was ever discussed anywhere, but was there any reason you opted for another specific number rather than generic language so that this problem didn't need to be fixed when RIT changes their mind about semesters again?

Something to the tune of 

`... attend at least the number of directorship meetings (including permanent and Ad-Hoc directorship meetings) equal to the number of weeks in each Academic Semester in which they are a member.`

Just a thought

csssuf:
Yeah, I thought about that, but it seemed like it made the language a little awkward. I'd be happy to change it if we can figure out a good way to phrase things.

zthart:
also maybe `in which -> during which`

so that it reads `... each Academic Semester during which they are an Active Member.`

zthart:
What about `... attend at least the number of directorship meetings ... equal to the length of each Academic Semester (in weeks) during which they are an active member.`

csssuf:
Yeah, that sounds pretty reasonable.

bmbowdish:
"Equal to the number of weeks of active membership" seems a little wordy. I think a number per semester would suffice.

bmbowdish:
Additionally, the wording would need to be EXTREMELY clear in that it would not mean "one meeting per week" but "one meeting per week by the end of the period"

csssuf:
Actually, I totally forgot the other reason I set it to a fixed number - implementing it in conditional is way easier as a fixed number. I guess I'd like to hear some more opinions on this one.

zthart:
The other benefit to being generic, is that it handles people leaving via a Leave of Absence, etc.

It may be harder to get right in conditional, but it’s far from impossible.

zthart:
Also, it is clear it doesn’t require attendance of one per week, nowhere in the text does it use the phrase “one meeting per week”. It says the number is equal to the number of weeks. 

See the comment on or current 10 weeks process

sgreene570:
Amendment passed

PR #128, opened 2018-02-11
Changes to Evaluations

bmbowdish:
Check one:

Summary of change(s):

* PR Directorship Created
* PR Directorship is given a budget of 5% of total funds
* Packet % increased to 75% 
* Packet is delayed a week. 
* Packets are no longer given out after week 2 and are due by the end of week 3. 
* Removal of Freshman Project
* Each semester the process repeats 

With this, some other changes will happen but they aren't constitutional in nature.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -683,7 +683,7 @@ This ruling may be appealed in the same manner as an Executive Board decision \r
 % ARTICLE VII - TEMPORARY AMENDMENTS
 \article{Temporary Amendments}
 In the case that an amendment should be temporary, the amendment's changes should be listed here. 
-Each of these amendments is given a designated period of time, and will cease to exist immedietely. 
+Each of these amendments is given a designated period of time, and will cease to exist immediately.
 If at any point interpretation of the changes needs to be made, the Executive Board will use the process defined in \ref{Judicial} to come to an agreement on how to proceed. 
 Occasionaly, Temporary Amendments may become permanent through a referendum, which is described in detail as part of each Temporary Amendment. 
 \asection{Changes to Evaluations}
@@ -709,7 +709,7 @@ Public Relations Director is added to \ref{Voting Members}, and is given the fol
 	\item To advertise any events run by Evaluations Director and/or Social Director(s) that are intended to be attended by RIT Students whom are not Computer Science House Members.  
 \end{enumerate}
 \asubsubsection{Evaluations and Social Directorship Changes}
-The Evalutations Director and Social Director(s) will have the following additional responsibilities during the described period. 
+The Evaluations Director and Social Director(s) will have the following additional responsibilities during the described period. 
 \begin{enumerate}
 	\item To preside over a number of events intended to produce interest in CSH, especially in RIT students during the begining weeks of each semester.
 \end{enumerate}
--- a/articles.tex
+++ b/articles.tex
@@ -680,4 +680,42 @@ The Judicial Board will be responsible for making all necessary inquiries into t
 The Judicial Board may present an official ruling following the Judicial Investigation.
 This ruling may be appealed in the same manner as an Executive Board decision \ref{Appeals}.
 
+% ARTICLE VII - TEMPORARY AMENDMENTS
+\article{Temporary Amendments}
+In the case that an amendment should be temporary, the amendment's changes should be listed here. 
+Each of these amendments is given a designated period of time, and will cease to exist immedietely. 
+If at any point interpretation of the changes needs to be made, the Executive Board will use the process defined in \ref{Judicial} to come to an agreement on how to proceed. 
+Occasionaly, Temporary Amendments may become permanent through a referendum, which is described in detail as part of each Temporary Amendment. 
+\asection{Changes to Evaluations}
+During the 2018-2019 period the following will supercede the text otherwise found in this document.
+\asubsection{Changes}
+\asubsubsection{Removal of Introductory Project}
+\ref{The Introductory Project} will no longer be a requirement as described in \ref{The Introductory Process}. 
+Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period. 
+\asubsubsection{Change of Packet Percentage}
+\ref{Creation of Accounts} currently stipulates that Introductory Members must acquire 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations. 
+During the described period, the percentage of required signatures in the Introductory Packet is increased to 75\%. 
+\asubsubsection{Changes to Packet Distribution}
+Introductory Members will recieve an \ref{The Introductory Packet} at the begining of the second week of each semester. 
+Any applicants who wish to become Introductory Members after the third week of each semester will recieve their packet during the begining of the second week of the following semester. 
+During the first, second, or third week of each semester, any applicant may have their application evaluated during a standard Evaluations Directorship Meeting or during either House Meeting or an Executive Board Meeting. 
+\asubsubsection{PR Director} 
+Public Relations Director is added to \ref{Voting Members}, and is given the following responsibilities during the described period.  
+\begin{enumerate}
+	\item To preside over Public Relations Directorship Meetings.
+	\item To exercise general supervision over the Public Relations operations.
+	\item To communicate clearly with the Rochester Institute of Technology's extended community.
+	\item To collaborate with the Social and Evaluations Directors to ensure that \ref{The Introductory Process} is directed correctly.
+	\item To advertise any events run by Evaluations Director and/or Social Director(s) that are intended to be attended by RIT Students whom are not Computer Science House Members.  
+\end{enumerate}
+\asubsubsection{Evaluations and Social Directorship Changes}
+The Evalutations Director and Social Director(s) will have the following additional responsibilities during the described period. 
+\begin{enumerate}
+	\item To preside over a number of events intended to produce interest in CSH, especially in RIT students during the begining weeks of each semester.
+\end{enumerate}
+\asubsubsection{Evaluation Period}
+During the described period, the \ref{The Introductory Process} will begin during the first week of each semester or when the Evaluations Director otherwise chooses to initiates it.
+The Executive Board may hold a Simple Majority vote to extend the Introductory Process or the Introductory Packet during the described period. 
+\subsubsection{Referendum}
+During the last house meeting of the second semester of the described period, a Two-Thirds vote is held, and if it passes, \ref{Changes to Evaluations} becomes permanent. 
 \end{document}
--- a/articles.tex
+++ b/articles.tex
@@ -683,39 +683,36 @@ This ruling may be appealed in the same manner as an Executive Board decision \r
 % ARTICLE VII - TEMPORARY AMENDMENTS
 \article{Temporary Amendments}
 In the case that an amendment should be temporary, the amendment's changes should be listed here. 
-Each of these amendments is given a designated period of time, and will cease to exist immediately.
+Each of these amendments is given a designated period of time, and will cease to exist immediately following the end of the period.
 If at any point interpretation of the changes needs to be made, the Executive Board will use the process defined in \ref{Judicial} to come to an agreement on how to proceed. 
-Occasionaly, Temporary Amendments may become permanent through a referendum, which is described in detail as part of each Temporary Amendment. 
+Occasionally, Temporary Amendments may become permanent through a referendum, which is described in detail as part of each Temporary Amendment. 
 \asection{Changes to Evaluations}
 During the 2018-2019 period the following will supercede the text otherwise found in this document.
 \asubsection{Changes}
 \asubsubsection{Removal of Introductory Project}
 \ref{The Introductory Project} will no longer be a requirement as described in \ref{The Introductory Process}. 
-Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period. 
+Additionally, \ref{Introductory Executive Board} will no longer be elected or perform any duties during the described period. 
 \asubsubsection{Change of Packet Percentage}
 \ref{Creation of Accounts} currently stipulates that Introductory Members must acquire 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations. 
-During the described period, the percentage of required signatures in the Introductory Packet is increased to 75\%. 
-\asubsubsection{Changes to Packet Distribution}
-Introductory Members will recieve an \ref{The Introductory Packet} at the begining of the second week of each semester. 
-Any applicants who wish to become Introductory Members after the third week of each semester will recieve their packet during the begining of the second week of the following semester. 
-During the first, second, or third week of each semester, any applicant may have their application evaluated during a standard Evaluations Directorship Meeting or during either House Meeting or an Executive Board Meeting. 
+During the described period, the percentage of required signatures in the Introductory Packet will be increased to 75\%. 
+\asubsubsection{Changes to Packet Times}
+Introductory Members will recieve \ref{The Introductory Packet} following the House Meeting or Evaluations Directorship meeting of the second week of each semester. 
+Each Introductory Member will submit their packet at the end of week 3. 
 \asubsubsection{PR Director} 
-Public Relations Director is added to \ref{Voting Members}, and is given the following responsibilities during the described period.  
+The Public Relations Director will be added to \ref{Voting Members}. 
+The Public Relations Director will be given half of Accumulated's budget (5\% of the total Directorships Budget.) 
+The Public Relations Director will be given the following responsibilities during the described period. 
 \begin{enumerate}
 	\item To preside over Public Relations Directorship Meetings.
-	\item To exercise general supervision over the Public Relations operations.
-	\item To communicate clearly with the Rochester Institute of Technology's extended community.
+	\item To exercise general supervision over Public Relations operations.
+	\item To communicate clearly with Rochester Institute of Technology's extended community.
 	\item To collaborate with the Social and Evaluations Directors to ensure that \ref{The Introductory Process} is directed correctly.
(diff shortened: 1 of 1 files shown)


articles.tex
```diff
@@ -680,4 +680,36 @@
...
+During the 2018-2019 period the following will supercede the text otherwise found in this document.
+\asubsection{Changes}
+\asubsubsection{Removal of Introductory Project}
+\ref{The Introductory Project} will no longer be a requirement as described in \ref{The Introductory Process}. 
+Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period. 
+\asubsubsection{Change of Packet Percentage}
+\ref{Creation of Accounts} currently stipulates that Introductory Members must acquire 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations. 
+During the described period, the percentage of required signatures in the Introductory Packet is increased to 75\%. 
+\asubsubsection{Changes to Packet Times}
+Introductory Members will recieve an \ref{The Introductory Packet} during the second week of each semester.
```
lueking: At what point during week 2 will the packet be handed out? This language allows packets to be handed out both at the start and the end of the week.

articles.tex
```diff
@@ -680,4 +680,36 @@
 The Judicial Board may present an official ruling following the Judicial Investigation.
 This ruling may be appealed in the same manner as an Executive Board decision \ref{Appeals}.
 
+% ARTICLE VII - TEMPORARY AMENDMENTS
+\article{Temporary Amendments}
+In the case that an amendment should be temporary, the amendment's changes should be listed here. 
+Each of these amendments is given a designated period of time, and will cease to exist immediately.
```
lueking: This sentence implies that the amendment expires both in the future and as soon as it comes into effect.

devinmatte:
I have an issue here with how we're doing applications now.
If we accept someone in week 4, and they can start getting attendance right away. But their 10 weeks is "10 weeks from the start of packet". That leaves them with a 20 week process.
Also I want to clarify that 10 weeks is 10 weeks from start of packet, or that 10 weeks total. Because packet starts on week 2. So that means "10 week" would be on week 11.

articles.tex
```diff
@@ -680,4 +680,36 @@
...
+If at any point interpretation of the changes needs to be made, the Executive Board will use the process defined in \ref{Judicial} to come to an agreement on how to proceed. 
+Occasionaly, Temporary Amendments may become permanent through a referendum, which is described in detail as part of each Temporary Amendment.
```
csssuf: `s/Occasionaly/Occasionally/`

articles.tex
SethGower: I would suggest, "Will expire after that time"

articles.tex
csssuf: This may be the same thing that @devinmatte mentioned in his comment, but I'd change this to "the second week after their application is accepted", or something to that effect - this fails to take into account anybody whose application process starts after the beginning of the semester.

articles.tex
```diff
@@ -680,4 +680,36 @@
...
+\asubsection{Changes}
+\asubsubsection{Removal of Introductory Project}
+\ref{The Introductory Project} will no longer be a requirement as described in \ref{The Introductory Process}. 
+Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period. 
+\asubsubsection{Change of Packet Percentage}
+\ref{Creation of Accounts} currently stipulates that Introductory Members must acquire 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations. 
+During the described period, the percentage of required signatures in the Introductory Packet is increased to 75\%. 
+\asubsubsection{Changes to Packet Times}
+Introductory Members will recieve an \ref{The Introductory Packet} during the second week of each semester. 
+Each Introductory Member finishes packet at the end of week 3.
```
SethGower: I think that we should specify a hard length of packet. With this wording, it starts somewhere in week 2 (I realize this is being discussed elsewhere also), and ends at the end of week 3. We should state that the packet will end 14 days after it starts.

bmbowdish: It begins during week 2, and ends at the beginning of week 4 (end of week 3 really.) 

This is intentional wording. It means a packet can be given out at any time during week 2, but always end by the end of week 3.

articles.tex
bmbowdish: This wording is intended to allow different scenarios to be handled in similar ways. 

Applicants accepted during the evals meeting of the second week would be given a packet and have until the end of the third week to finish it. Applicants before the 2nd week would begin during the beginning of the week. 

Any applicants after week 2 will not receive a packet until the following semester.

articles.tex
```diff
@@ -680,4 +680,37 @@
...
+	\item To preside over Public Relations Directorship Meetings.
+	\item To exercise general supervision over the Public Relations operations.
+	\item To communicate clearly with the Rochester Institute of Technology's extended community.
+	\item To collaborate with the Social and Evaluations Directors to ensure that \ref{The Introductory Process} is directed correctly.
+	\item To advertise any events run by Evaluations Director and/or Social Director(s) that are intended to be attended by RIT Students whom are not Computer Science House Members.  
+\end{enumerate}
+\asubsubsection{Evaluation Period}
+During the described period, the \ref{The Introductory Process} will begin during the first week of each semester.
+The Executive Board may hold a Simple Majority vote to extend the Introductory Process or the Introductory Packet during the described period.
+Additionally, \ref{Introductory Evaluation} occurs during between the sixth and tenth week of each semester.
```
csssuf: `s/during //`

articles.tex
```diff
@@ -680,4 +680,37 @@
...
+	\item To preside over Public Relations Directorship Meetings.
+	\item To exercise general supervision over the Public Relations operations.
+	\item To communicate clearly with the Rochester Institute of Technology's extended community.
+	\item To collaborate with the Social and Evaluations Directors to ensure that \ref{The Introductory Process} is directed correctly.
+	\item To advertise any events run by Evaluations Director and/or Social Director(s) that are intended to be attended by RIT Students whom are not Computer Science House Members.  
+\end{enumerate}
+\asubsubsection{Evaluation Period}
+During the described period, the \ref{The Introductory Process} will begin during the first week of each semester.
+The Executive Board may hold a Simple Majority vote to extend the Introductory Process or the Introductory Packet during the described period.
+Additionally, \ref{Introductory Evaluation} occurs between the sixth and tenth week of each semester.
```
bmbowdish: @devinmatte Is this what you wanted?

devinmatte: Yes that language is much more clear.
This new system may create synthetic "20 week" evaluations, but I feel that an implementation error we'll have to just account to in trying this system

bmbowdish: I understand the concern, but there are members who fail spring evals and experience exactly what you're concerned about currently.

sgreene570:
@bmbowdish can you squash your changes into one commit?

articles.tex
lueking: If we are trying to drive recruitment over the course of the week, this doesn't really work. An applicant would have to wait for an evals meeting to start packet. So if they show up to an event after an evals meeting and want to join, they can't get a packet. 
I would really like a more specific wording for this section.

articles.tex
```diff
@@ -680,4 +680,37 @@
...
+	\item To communicate clearly with the Rochester Institute of Technology's extended community.
+	\item To collaborate with the Social and Evaluations Directors to ensure that \ref{The Introductory Process} is directed correctly.
+	\item To advertise any events run by Evaluations Director and/or Social Director(s) that are intended to be attended by RIT Students whom are not Computer Science House Members.  
+\end{enumerate}
+\asubsubsection{Evaluation Period}
+During the described period, the \ref{The Introductory Process} will begin during the first week of each semester.
+The Executive Board may hold a Simple Majority vote to extend the Introductory Process or the Introductory Packet during the described period.
+Additionally, \ref{Introductory Evaluation} occurs between the sixth and tenth week of each semester. 
+\subsubsection{Referendum}
+During the last house meeting of the second semester of the described period, a Two-Thirds vote is held, and if it passes, \ref{Changes to Evaluations} becomes permanent.
```
sgreene570: This sentence could use some work.  `Becomes` should probably be `becomes and we should probably avoid comma splices.

csssuf: I think `Becomes` is correct here; "Changes to Evaluations" is a single section.

bmbowdish: So I shouldn't change it?

sgreene570: Yea it is correct as is, my mistake.

articles.tex
```diff
@@ -680,4 +680,39 @@
...
+Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period. 
+\asubsubsection{Change of Packet Percentage}
+\ref{Creation of Accounts} currently stipulates that Introductory Members must acquire 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations. 
+During the described period, the percentage of required signatures in the Introductory Packet is increased to 75\%. 
+\asubsubsection{Changes to Packet Times}
+Introductory Members will recieve \ref{The Introductory Packet} following the House Meeting or Evaluations Directorship meeting of the second week of each semester. 
+Each Introductory Member will submit their packet at the end of week 3. 
+\asubsubsection{PR Director} 
+The Public Relations Director is added to \ref{Voting Members}. 
+The Public Relations Director will be given half of Accumulated's budget (5\% of the total Directorships Budget.)
```
bmbowdish: @devinmatte and I talked about giving PR a budget.

bmbowdish:
I just gave PR

bmbowdish:
Alright ladies and gents. This is looking like it is approaching the final version of this amendment. Does anyone else have concerns?

SethGower:
I think that it looks good. Just what @sgreene570 said earlier, just squash the commits.

grahams:
My main concern is that you are lumping three seemingly unrelated things into one amendment.  As far as I can tell you are:

1. Creating the concept of Temporary Amendments
1. Murdering the Freshman Project/Board
1. Creating a new directorship for PR.

I don't understand the rationale.  You are basically asking voters to either pass/fail the entire thing or swallow something they may not like to get something they do.

bmbowdish:
These 3 things are all related.

The removal of freshman project and changes to packet requires the two other portions to be run properly.

grahams:
@bmbowdish Do you have an explanation of this somewhere, because that is not at all evident from the PR.

bmbowdish:
There have been several presentations given with notes taken multiple times.

grahams:
I think when you say "These 3 things are all related." it implies they are all inter-required.  But what I really think you mean is: 

1. "I think we need to remove the freshman project"
1. "as part of that, I think we need to replace it with a PR director to outreach, recruit new members, etc"
1. "but some people are uncomfortable with that idea and consider it a big change, so lets create the idea of temporary amendments."

Just because you need the latter steps to get to the former doesn't really mean they are interdependent.  You could have temporary amendments or a PR director and they could be great for the house without any changes to evals.  And you could have temp. amendments without the other two and it could still be great.

I really like the idea of temporary (or, even better, "experimental") amendments, I'm relatively indifferent to the idea of a PR directorship (though if there are an even number of exec board members it can make things complicated) and i'm _vaguely_ opposed to the removal of the charity project.  

If I had a vote, which, of course, I do not, I would (presently) definitely vote this amendment down.  

But I can also imagine a world where I voted Yes for three discrete amendments.  But I would definitely need more persuasion on the latter two.  

That's my two cents as an old man.

bmbowdish:
So, I’ll comment on the thing you pointed out your being vaguely against. 

This year the entire freshman class EBoard failed fall evals. The freshman project for the last 3-4 years has been pretty negative to the culture of the intro class and the feeling of almost all of floor is that it needs to go. 

Also, Social is currently a single Director, and has been for 2 years. We like not having dual directors and so during this temp year having a new person has the same effect as there’s been for 40 years.

bmbowdish:
I appreciate old man comments, I just think that this is a pretty aggressive amendment, so I understand why some alumni look at it and aren’t 100% behind it.

articles.tex
```diff
@@ -680,4 +680,39 @@
...
+In the case that an amendment should be temporary, the amendment's changes should be listed here. 
+Each of these amendments is given a designated period of time, and will cease to exist immediately following the end of the period.
+If at any point interpretation of the changes needs to be made, the Executive Board will use the process defined in \ref{Judicial} to come to an agreement on how to proceed. 
+Occasionally, Temporary Amendments may become permanent through a referendum, which is described in detail as part of each Temporary Amendment. 
+\asection{Changes to Evaluations}
+During the 2018-2019 period the following will supercede the text otherwise found in this document.
+\asubsection{Changes}
+\asubsubsection{Removal of Introductory Project}
+\ref{The Introductory Project} will no longer be a requirement as described in \ref{The Introductory Process}. 
+Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period.
```
stevenmirabito: Remove `also` on this line, you're already saying `Additionally`

articles.tex
```diff
@@ -680,4 +680,39 @@
...
+Occasionally, Temporary Amendments may become permanent through a referendum, which is described in detail as part of each Temporary Amendment. 
+\asection{Changes to Evaluations}
+During the 2018-2019 period the following will supercede the text otherwise found in this document.
+\asubsection{Changes}
+\asubsubsection{Removal of Introductory Project}
+\ref{The Introductory Project} will no longer be a requirement as described in \ref{The Introductory Process}. 
+Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period. 
+\asubsubsection{Change of Packet Percentage}
+\ref{Creation of Accounts} currently stipulates that Introductory Members must acquire 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations. 
+During the described period, the percentage of required signatures in the Introductory Packet is increased to 75\%.
```
stevenmirabito: Inconsistent language. Stick with `will` (as used above, e.g. `will no longer`) or `is` (as used here, e.g. `is increased`)

articles.tex
```diff
@@ -680,4 +680,39 @@
...
+During the 2018-2019 period the following will supercede the text otherwise found in this document.
+\asubsection{Changes}
+\asubsubsection{Removal of Introductory Project}
+\ref{The Introductory Project} will no longer be a requirement as described in \ref{The Introductory Process}. 
+Additionally, \ref{Introductory Executive Board} will also no longer be elected or perform any duties during the described period. 
+\asubsubsection{Change of Packet Percentage}
+\ref{Creation of Accounts} currently stipulates that Introductory Members must acquire 60\% (rounded up to the nearest whole person) of required signatures (excluding those of Resident members who have not passed a Membership Evaluation) in the Introductory Packet or successfully complete Introductory Evaluations. 
+During the described period, the percentage of required signatures in the Introductory Packet is increased to 75\%. 
+\asubsubsection{Changes to Packet Times}
+Introductory Members will recieve \ref{The Introductory Packet} following the House Meeting or Evaluations Directorship meeting of the second week of each semester.
```
stevenmirabito: So some intro members will have a packet a few days longer than others, depending on if they start at house meeting or evals?

articles.tex
stevenmirabito: Why say `given half of Accumulated's budget` instead of just redefining the percentages here with this taken into account?

articles.tex
```diff
@@ -680,4 +680,39 @@
...
+\asubsubsection{Changes to Packet Times}
+Introductory Members will recieve \ref{The Introductory Packet} following the House Meeting or Evaluations Directorship meeting of the second week of each semester. 
+Each Introductory Member will submit their packet at the end of week 3. 
+\asubsubsection{PR Director} 
+The Public Relations Director is added to \ref{Voting Members}. 
+The Public Relations Director will be given half of Accumulated's budget (5\% of the total Directorships Budget.) 
+The Public Relations Director is given the following responsibilities during the described period. 
+\begin{enumerate}
+	\item To preside over Public Relations Directorship Meetings.
+	\item To exercise general supervision over the Public Relations operations.
```
stevenmirabito: I don't think this line is necessary. This section describes what "Public Relations operations" are. If we want to keep this I would suggest a wording of `supervision over Public Relations operations` (strike the `the`).

articles.tex
```diff
@@ -680,4 +680,39 @@
...
+	\item To communicate clearly with the Rochester Institute of Technology's extended community.
```
stevenmirabito: Similarly, strike `the`

articles.tex
bmbowdish: Some, not many, would have 1-3 less days than the rest of their peers. During this period people will know this and the applicants will still be able to do packet talks. 1-3 days isn't a concern for me, and I think we're still being pretty fair.

stevenmirabito: Yep, I don't think it's a huge deal either. Just wanted to make sure that was intentional.

articles.tex
bmbowdish: Are you sure?

articles.tex
```diff
@@ -680,4 +680,39 @@
 The Judicial Board may present an official ruling following the Judicial Investigation.
 This ruling may be appealed in the same manner as an Executive Board decision \ref{Appeals}.
 
+% ARTICLE VII - TEMPORARY AMENDMENTS
+\article{Temporary Amendments}
+In the case that an amendment should be temporary, the amendment's changes should be listed here. 
+Each of these amendments is given a designated period of time, and will cease to exist immediately following the end of the period.
```
momohossain: Replace ```is``` with 'are'.  "```Each of these amendments``` are ```given a...```"

bmbowdish: I'll change it once the amendment passes. That way everyone can vote yes to the same wording. Its a non-semantic change but ya know.

nicholastmosher: I believe it's actually correct as-is. After saying "each" you're referring to the singular item not the plural.

bmbowdish:
🥅 !

PR #129, opened 2018-02-13
Moving Housing Point Allocation to Spring Evals

devinmatte:
Check one:

Summary of change(s):
Removing the need for a Housing Points Form and moving the allocation of Housing Points to Spring evals.
1. To encourage upperclassmen to try on Spring Evals
2. To remove a need to just distribute points based on the filling out of a from


bmbowdish:
I don’t like this. No one should be punished for needing to take some time away.

bmbowdish:
Why not one housing point?

devinmatte:
@bmbowdish So I will definitely consider that following a discussion at House Meeting on the Amendment. I think that's something that can be determined there

PR #132, opened 2018-09-23
Update packet signature requirements

RamZallan:
Check one:

Summary of change(s):

- Requires signature of **all** active members (both on- and off-floor)
- Lowers requirement of alumni (non-active members), advisors, and honorary members to 10, since off-floors now count under the active member category

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -170,9 +170,8 @@ The advisors assist in organizing the participants at the beginning of the Intro
 \bsubsection{The Introductory Packet}
 Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain the following:
 \begin{itemize}
-	\item A signature from all Active/Introductory Resident members, each Executive Board member, and fifteen (15) of any combination of the following:
+	\item A signature from all Active members, all Introductory Resident members, each Executive Board member, and ten (10) of any combination of the following:
 	\begin{itemize}
-		\item Non-Resident members who have passed a Membership Evaluation
 		\item Alumni members (not including Non-Resident Executive Board members)
 		\item Honorary members
 		\item Advisory members
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -170,9 +170,9 @@ The advisors assist in organizing the participants at the beginning of the Intro
 \bsubsection{The Introductory Packet}
 Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain the following:
 \begin{itemize}
-	\item A signature from all Active members, all Introductory Resident members, each Executive Board member, and ten (10) of any combination of the following:
+	\item A signature from all Active members, all Introductory Resident members, and ten (10) of any combination of the following:
 	\begin{itemize}
-		\item Alumni members (not including Non-Resident Executive Board members)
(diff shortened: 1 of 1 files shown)


bmbowdish:
Hmm. 10 is a lot of alumni/advisors/honorary members. I don't know if they should even be a requirement. 

I don't like this change in general, but if floor wants it, that would be a concern of mine.

zedoax:
One thing I'd like to see out of this amendment is a more mitigatory thing.  There are a lot of off floor upperclassmen who either have the intent to enter alumni status or know they will not have time to visit floor.  I believe this change should include a 'participating off floor' definition, and the default status for off floors is opted out.  By this I mean Off Floor active members that choose to participate in packet will matter for the upperclassmen percent.  That way members who do not intend to be actively involved in packet will not put more strain on the evals director by having to remove them on a case by case basis.  This may also be functionality that could be added similar to being marked 'active' in conditional

RamZallan:
@zedoax 
I'll respond to a couple of your points:

>There are a lot of off floor upperclassmen who either have the intent to enter alumni status

That should be something taken up with the evals or financial director, and those people who can't come around because they don't want to be active can be discounted from packet totals, imo.

>...or know they will not have time to visit floor

IMO, people who are in this position shouldn't be active. They still have the requirements to meet the 15 directorship meetings every semester, and come to house meeting every Sunday, so they should be on floor enough to sign some packets. I think that any person who can't reasonably come to floor once a week probably doesn't need to be active, since they'll most likely find it hard to meet the other requirements, ignoring signing freshman packets.

If someone doesn't want to be involved with packet, they shouldn't be active, I think. Someone who pays dues not only has all the **benefits** of being an active member, but also all the **responsibilities**.

zedoax:
@RamZallan 
I do agree with your points.  However, that is the case with quite a few active members.  In an ideal scenario active off floor members would be on floor and participating in packet.  However, the reality is this is the situation with many off floor members.  It doesn't seem fair to eboard imo where they will have to decide towards the end of packet if someone has met arbitrary requirements (which afaik is up to eval's discretion) for being active on packet.  I see that becoming a point of contention, and would like to see that mitigated prior to voting yes.  Without a system as such, I do not believe we define what constitutes being 'active' on packet, and as such would like to see those defined for all members if we are to leave judgement of removal of packets up to the evals director.

RamZallan:
Just a note that the last commit removes the mention of eboard members, because currently all eboard members are required to be active. This made it redundant to mention all active members _and_ all eboard members. No semantics of that were actually changed.

RamZallan:
@zedoax agree to disagree, I guess. I don't see it as being too much on eboard to decide that, especially since it's a total hypothetical about some member who decides to be active (i.e. pay dues, go to directorship meetings, go to house meeting, etc.) but for some reason not sign packets. We don't have a definition anywhere because that sort of thing is really fluid and can change year-to-year. If we have a year of less-than-stellar applicants, maybe all active members will sign less packets compared to prior years, or maybe we'll have years where everyone signs the majority of packets. That's all neither here nor there and I don't think it's our place to define what a "good" signature number is.

bmbowdish:
big sad

PR #133, opened 2018-09-23
Remove 'essay' requirement from intro packet

RamZallan:
Check one:

Summary of change(s):
Removes the "essay" requirements from the intro packet. It's antiquated with no real reason to include it that I've heard other than to have a very low hurdle for freshmen. This year we had someone hit 100% on packet but "fail" packet because they didn't submit the essay. IMO, if house decides someone is good enough to get 100, that's more than enough proof that them looking at the Calendar and pubsite doesn't change anything.


bmbowdish:
I am totally on board.

PR #138, opened 2018-11-07
Grammar.

saucetray:
Check one:

Summary of change(s):
That comma shouldn't be there.


gambogi:
This phrasing still feels awkward. 
Should be: 
> The chairperson of the vote and two randomly selected members are Vote Counters
Or similar
Also this change is semantic! You lost “randomly”

PR #140, opened 2018-12-10
Updating How Bad Standing Votes Work

saucetray:
Check one:

Summary of change(s):
The way this is written is so bad. Theoretically, people voted to good standing should now be active and given off floor status which is just... bad.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -184,7 +184,7 @@ Active Membership is open to all students currently enrolled at the Rochester In
 Qualifying members may self-select into Active Membership by paying dues to the Financial Director and notifying the Evaluations Director.
 \\*\\*
 Any Alumni Member in bad standing (\ref{Alumni Membership Selection}) may become an Active Member by notifying the Evaluations Director of their intent to participate, who will bring them up for vote at the next Evaluations meeting.
-If a majority is in favor of the return, the Alumni is removed from bad standing.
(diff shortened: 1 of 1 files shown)


mxmeinhold:
Looks good.
One question, why remove the requirement to announce it?

saucetray:
Is it ever announced now?

mxmeinhold:
Hmm. I don’t feel a need to announce it so long as it’s included in notes I suppose.

articles.tex
```diff
@@ -184,8 +184,7 @@
 Qualifying members may self-select into Active Membership by paying dues to the Financial Director and notifying the Evaluations Director.
 \\*\\*
 Any Alumni Member in bad standing (\ref{Alumni Membership Selection}) may become an Active Member by notifying the Evaluations Director of their intent to participate, who will bring them up for vote at the next Evaluations meeting.
```
csssuf: This line should be dropped as well.

articles.tex
```diff
@@ -184,8 +184,7 @@
...
-If a majority is in favor of the return, the Alumni is reinstated as an Active Member with Off-Floor status effective immediately.
-The decision will be announced at the following House meeting.
+If a majority is in favor of the return, the Alumni is removed from bad standing.
```
csssuf:
```suggestion
If a majority is in favor of the return, the Alumni is considered to be in good standing.
```

PR #143, opened 2019-01-17
Remove the Temporary Amendment

torirossini:
The temporary amendment was taken apart bit by bit until all the useful parts of it were gone. All that remained was the ability to make temporary amendment, made the packet percentage to get an account higher, and disallowed intro members from reapplying/applying past the first week of each semester. I'd like to propose we don't cause this semester's intro members to go through a half-deconstructed process which we will get rid of entirely at the last HM of the semester. Just end it now.

Check one:
- [X ] Semantic Change: something about the meaning of the text is different

Summary of change(s):


bmbowdish:
It’s been dead since week 2.

PR #148, opened 2019-02-18
Elections Must Be Held for the Executive Board

saucetray:
Check one:

Summary of change(s):

We should have a process which requires elections to be held once someone resigns instead of having a possible interim director forever at the discretion of the Executive Board.

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -534,7 +534,7 @@ The duties and responsibilities of a vacated office are assumed by the Chairman 
 The vote of a vacated Executive Board position shall be cast as an abstention in all Executive Board matters where this vote is required to be cast.
 
 \asection{Delaying an Election}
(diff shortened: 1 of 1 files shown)


saucetray:
Let me know if there are any issues.

articles.tex
```diff
@@ -526,7 +526,9 @@
 \asection{Resignations}
 An Executive Board Member may resign their position by submitting in writing the reason for resignation to the House Chairman.
 The resignation will be read at the following House Meeting and become effective at that time.
-The office will become vacant and the selection process for a new member as described in section \ref{Appointment of an Interim Director} will begin at that time.
+The office will become vacant and the selection process for a new member as described in section \ref{Appointment of an Interim Director} will begin at that time.
```
mxmeinhold: Oh, this is showing up in the diff because there’s a space at the end of the line I believe.

saucetray: We can just make a nonsemantic after ballots if it works

PR #150, opened 2019-02-25
Outdated Housing Points Reference

saucetray:
Check one:

Summary of change(s):
If I'm reading this correctly... this is simply referring to winter evals that doesn't exist anymore.

### DIFF
--- a/bylaws.tex
+++ b/bylaws.tex
@@ -299,7 +299,7 @@ They still have access to all other privileges associated with their membership,
 \bsection{Housing Priority System}
 The Housing Priority System is a means for determining the priority a House member has in a Housing issue such as Single Room Assignment or the Assignment of Available Housing.
 The member with top priority is the member with on-floor status and the most Housing Priority Points.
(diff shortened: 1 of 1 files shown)


bylaws.tex
```diff
@@ -299,7 +299,7 @@
 \bsection{Housing Priority System}
 The Housing Priority System is a means for determining the priority a House member has in a Housing issue such as Single Room Assignment or the Assignment of Available Housing.
 The member with top priority is the member with on-floor status and the most Housing Priority Points.
-Housing points are accumulated during the Housing Evaluation described in \ref{Housing Evaluation}.
+Housing points are accumulated by the end of the year in accordance with the Evaluations Director. \ref{Housing Points}.
```
devinmatte: > accordance with the Evaluations Director
Doesn't feel right. I don't even think that's proper grammar.
I agree this should be updated, but I'm not sure that's the correct way to say it

saucetray: Let's yeet all of this and just make housing points attainable at the end of spring evals

saucetray:
Closed for now, will revisit next year.

PR #154, opened 2019-03-01
Remove the notarize reference

mxmeinhold:
We don't require this per the constitution

Check one:

Summary of change(s): Per the constitution, we no longer need to print, sign, notarize, and place the document in the cabinet. But it's a file change, so it's semantic.


saucetray:
Uhm... that is the most ignored thing I’ve ever seen in this document.

saucetray:
Is this... semantic.

bmbowdish:
Sort of.

saucetray:
Yikes

PR #160, opened 2019-04-09
Replaced all instances of Chairman with Chairperson

andreweggleston:
Check one:

Summary of change(s):
Makes chairman inclusive!


saucetray:
Why not rename to president and put it through house

saucetray:
I am hosting an informal poll tomorrow at house meeting to determine the following way to define the chairman:

*Chairman
*Chairwoman
*Chairperson
*Chair
* [Whatever Anyone Else Suggests]

andreweggleston:
This looks reaaally stupid with Chair. People seemed to like it but also didn't actually read it....

devinmatte:
Chair would be okay if we did not use chair in other instances (chair of a vote). It becomes more confusing in text.
It doesn't change what we call them (chairman, Chairwoman, chairperson) but I think the text should be chairperson as to not conflict with the existing uses of chair in the document

mxmeinhold:
Resolved by #172

PR #162, opened 2019-04-21
Simplify the requirements for nominations

mxmeinhold:
By current practice, we don't restrict nomination only to qualified
members, so we should update that, so we can bring up candidates and
determine whether their merit should override the qualifications of a
seat.

Check one:

Summary of change(s):

This is primarily to update to current practice. If we are going to bring up a candidate who isn't qualified and consider the possibility of an override of the qualifications, we allow them to be nominated before we do discussion and voting. This means we can bring up overrides at a House Meeting before ballots are distributed.

This change removes the requirement to be qualified from nominations and places it instead with the inclusion in ballots and the determination of winners.


bmbowdish:
I worry this complicates the document while simplifying the process. This change seems like something someone would miss while reading the document for the first time.

saucetray:
People miss ebeeyrhint

bmbowdish:
If the goal is to improve the long term health of house, which is what it should always be imo, making small amendments people are going to forget about is a waste of time and damages the structure of the organization.

PR #163, opened 2019-04-21
Automatically pass All Green at Spring Evals

mxmeinhold:
Check one:

Summary of change(s):

All members who have completed all of their requirements will automatically pass their Membership Evaluation.

This stems from recent conversations, where it has sounded as though we should never not be passing a member who has objectively completed their requirements.


bmbowdish:
Giving members less control over evaluations in general is a bad idea in my opinion.

saucetray:
Well then we make spring evals subjective. There's no reason an objective evaluations would fail someone with all greens. That's insane. Then we can use objectives for people who are not all green and any applicable excuses to pass/fail/conditional said member.

mxmeinhold:
Rebased to resolve merge conflicts

PR #164, opened 2019-04-22
Chairman -> Chair

saucetray:
Check one:

Summary of change(s):

### DIFF
--- a/articles.tex
+++ b/articles.tex
@@ -173,7 +173,7 @@ The introductory evaluation process is described in \ref{Introductory Evaluation
 \asubsection{Introductory Membership Leave of Absence}
 After completion of the Introductory Packet, an Introductory member may request a leave of absence through the process described in \ref{Leave of Absence}.
 \asubsection{Introductory Membership Resignations}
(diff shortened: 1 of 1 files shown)


articles.tex
```diff
@@ -173,7 +173,7 @@
 \asubsection{Introductory Membership Leave of Absence}
 After completion of the Introductory Packet, an Introductory member may request a leave of absence through the process described in \ref{Leave of Absence}.
 \asubsection{Introductory Membership Resignations}
-Introductory members may resign by submitting the request for termination of membership to the Chairman, or Evaluations Director in writing before the completion of the introductory process.
+Introductory members may resign by submitting the request for termination of membership to the Chari, or Evaluations Director in writing before the completion of the introductory process.
```
devinmatte: Chari

mxmeinhold:
@devinmatte We use `chair` and `chairperson` as in chairperson of the vote once each. This changes both of those to `chairperson of the vote` To resolve that conflict.

I do agree that reading the uses of `Chair` feels really out of place. I would prefer us to use `Chairperson` and `chair of the vote` instead.

PR #174, opened 2019-10-06
Rm housing evals

mxmeinhold:
Check one:

Summary of change(s):

Deletes bad references to Housing Evals, makes housing points automatic. Also updates constitution to current process.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -673,7 +673,7 @@ They still have access to all other privileges associated with their membership,
 \asubsubsection{Housing Priority System}
 The Housing Priority System is a means for determining the priority a House member has in a Housing issue such as Single Room Assignment or the Assignment of Available Housing.
 The member with top priority is the member with on-floor status and the most Housing Priority Points.
-Housing Priority Points are accumulated once per Operating Session at the conclusion of the Membership Evaluation. Each Active Member with On-Floor Status who passes the Membership Evaluation is granted one (1) Housing Priority Point.
+Housing Priority Points are accumulated once per Operating Session at the conclusion of the Membership Evaluation. Each Active Member with On-Floor Status who passes the Membership Evaluation is granted two (2) Housing Priority Points.
 \\* \\*
 In the event of a tie, the members will be approached simultaneously and if they are unable to decide fairly between themselves, the assignment of priority will be made by random selection of the tied members.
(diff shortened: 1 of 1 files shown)


saucetray:
LGTM 🎆 🍾

nicholastmosher:
In theory this should fix the state of the points system, but in practice this will only work starting with incoming freshmen. I think that housing points should not need to be "accumulated", but instead that they should simply be a property of your membership that is derived from the number of years you've been an active member. This would make a difference in a case like myself, where I have already "missed" housing points in the past.

* Under this proposed amendment as-is, I would - from this point forward - always receive my 2 housing points each year. However, since I "missed" housing points in the past under the old system, I still only have 4 housing points.
* Under the "derived points" system, the number of points I have is simply calculated from the number of years I've been a member (i.e. years since passing my first spring evals). If the calculation is simply "number of years as a member * 2", then I would have (8 points? 10 points? depending on whether points get added at the beginning or end of the year)

I'm in favor of a "derived points" system because it eliminates the edge case where people missed housing points from the old (and broken) points system.

constitution.tex
```diff
@@ -673,11 +673,11 @@
...
 The member with top priority is the member with on-floor status and the most Housing Priority Points.
-Housing points are accumulated during the Housing Evaluation described in \ref{Housing Evaluation}.
+Housing Priority Points are accumulated once per Operating Session at the conclusion of the Membership Evaluation. Each Active Member with On-Floor Status who passes the Membership Evaluation is granted two (2) Housing Priority Points.
 \\* \\*
 In the event of a tie, the members will be approached simultaneously and if they are unable to decide fairly between themselves, the assignment of priority will be made by random selection of the tied members.
 \asubsubsubsection{Single Room Assignments}
-When a single room on the House becomes available it is offered to the member who carries the highest Housing Priority as defined in By-Law V, Section H.
+When a single room on the House becomes available it is offered to the member who carries the highest Housing Priority as defined in \ref{Housing Priority System}.
```
saucetray: What if two people have the same housing priority

mxmeinhold: Copy the line about ties from just above?

mxmeinhold:
Handed out: 61
Yes: 58
No: 2
Abstain: 1

PR #178, opened 2019-11-03
Correct Makefile to handle references

mxmeinhold:
Check one:

Summary of change(s):
The makefile needs to build twice so that previous references are understood.


saucetray:
LGTM

PR #182, opened 2019-11-05
Don't elect OpComm, Ad Hocs, or the Secretary

mxmeinhold:
Check one:

Summary of change(s):
#181 and #148 would require an election in the event of the resignation of the OpComm Director, the Secretary, or any AdHoc Director. This requires their replacement to be selected via the process we usually use to select that position.


saucetray:
Should we add impeachments

mxmeinhold:
Passes
75 Yes
2 No
0 Abstain

78 Handed out

mxmeinhold:
@devinmatte @saucetray Could I get a review and merge?

PR #183, opened 2020-04-09
Remove outdated 'Housing Points' and cleanup Housing Priority

mxmeinhold:
Check one:

Summary of change(s):

We use the 'Housing Priority System' and 'Housing Priority Points'. We
don't need to also define 'Housing Points' in a different place, with a
conflicting definition.

This PR brings the constitution in line with current practice, and makes the constitution less confusing and self contradictory.


mxmeinhold:
Ballots required for vote: 38
Ballots Cast: 44
Yes: 43
No: 1
Result: Passes

devinmatte (approved):

PR #193, opened 2021-02-07
Return auto pass to the written constitution

mxmeinhold:
Check one:

Summary of change(s):

We originally introduced auto-pass in #163, which was brought up to vote, and passed.
It seems that when the documents were merged in #172, this addition was lost.

This PR introduces the same wording, in the same sub section. This is exactly the same change as #163, and I believe current practice as of last year.

The original wording of #163 was:
```
All members who have completed all of their requirements will automatically pass their Membership Evaluation.
```


mxmeinhold:
### Ballot results:

Quorum: 40.66
Votes submitted: 50
Yes Votes Required to Pass: 25

Yes: 45
No: 5
Abstain: 0

Result: The amendment passes

saucetray (approved):
LGTM

PR #197, opened 2021-09-21
Added Polygamy/Polyamory Directorship

Cbturtle2:
Check one:

Summary of change(s):
Added Polygamy/Polyamory Directorship


bmbowdish:
Is RIT cool with CSH having someone who's in charge of managing personal relationships like this? In my time this would have definitely been a little controversial for more than that reason alone, but I understand that things change and evolve with time.

PR #202, opened 2021-11-21
Updated introductory process timeline to new proposal

crigney3:
Check one:

Summary of change(s):
These changes are a proposal to bring the introductory process timeline into the new system proposed throughout discussion this year. Instead of a ten week process, the process would now be six weeks long, and could occur twice in a semester. See https://discourse.csh.rit.edu/t/proposing-possible-changes-to-the-intro-process/130335/11 for further discussion on this topic.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -520,7 +520,8 @@ Funds allocated for a project not spent by the end of the Standard Operating Ses
 \asubsubsection{The Introductory Process}
 The Introductory Process is designed to provide an easy means for Introductory Members to meet existing House members, learn House history, demonstrate their involvement potential to House, and allow existing House members to evaluate them for Active Membership.
 \asubsubsubsection{The Evaluation Period}
-The Process will occur either once or twice a semester, at the discretion of the Evaluations Director, and will last for six (6) weeks.
+The Process will occur either once or twice per semester, at the discretion of the Evaluations Director, and will last for six (6) weeks.
+The Process shall be initiated by the Evaluations Director during either the first or second week of each semester.
 The Executive Board may hold a closed ballot Simple Majority vote to extend the Introductory Process or the Introductory Packet.
 If an Introductory Process is extended, the Introductory Evaluation shall occur at the termination of the extended Process, which may be outside of the period defined in \ref{Introductory Evaluation}.
 If deemed necessary by the Evaluations Director, the second intro process must begin within nine (9) days of the first process ending.
--- a/constitution.tex
+++ b/constitution.tex
@@ -520,11 +520,11 @@ Funds allocated for a project not spent by the end of the Standard Operating Ses
 \asubsubsection{The Introductory Process}
 The Introductory Process is designed to provide an easy means for Introductory Members to meet existing House members, learn House history, demonstrate their involvement potential to House, and allow existing House members to evaluate them for Active Membership.
 \asubsubsubsection{The Evaluation Period}
-The Process will occur either once or twice per semester, at the discretion of the Evaluations Director, or by an Executive Board simple majority vote, and will last for six (6) weeks.
+The Process will occur either once or twice per semester, and will last for six (6) weeks.
 The Process shall be initiated by the Evaluations Director during either the first or second week of each semester.
 The Executive Board may hold a closed ballot Simple Majority vote to extend the Introductory Process or the Introductory Packet.
 If an Introductory Process is extended, the Introductory Evaluation shall occur at the termination of the extended Process, which may be outside of the period defined in \ref{Introductory Evaluation}.
-If deemed necessary by the Evaluations Director, the second intro process must begin within nine (9) days of the first process ending.
+If deemed necessary by the Evaluations Director, or by an Executive Board simple majority vote, the second intro process must begin within nine (9) days of the first process ending.
 \asubsubsubsection{The Introductory Packet}
 Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain a signature from all Active members, all Introductory Resident members, and ten (10) of any combination of the following:
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


constitution.tex
```diff
@@ -544,7 +545,8 @@
 
 \asubsubsubsection{Introductory Evaluation}
 The Introductory Evaluation is the process by which House chooses which Introductory Members to extend an offer of full Resident or Non-Resident Membership.
-Occurs between the sixth and tenth week of each semester.
+Occurs during the sixth week of each semester. If the evaluations director chooses to hold a second Introductory Process, the Introductory Evaluation will also occur during the twelfth week of each semester.
```
mxmeinhold, line 548:
```suggestion
Occurs at the conclusion of the sixth week of the Process.
```
Since the process may start the first or second week, maybe this should be tied to the length of the process, rather than the date?

mxmeinhold: Also, each sentence gets its own line

constitution.tex
```diff
@@ -520,9 +520,10 @@
...
 \asubsubsubsection{The Evaluation Period}
-The Process will begin between the first and fourth week of each semester, then last between six (6) and ten (10) weeks, starting when the Evaluations Director initiates it.
+The Process will begin between the first and second week of each semester, then last six (6) weeks, starting when the Evaluations Director initiates it.
 The Executive Board may hold a closed ballot Simple Majority vote to extend the Introductory Process or the Introductory Packet.
 If an Introductory Process is extended, the Introductory Evaluation shall occur at the termination of the extended Process, which may be outside of the period defined in \ref{Introductory Evaluation}.
+The Introductory Process may be held twice a semester at the discretion of the evaluations director. If the evaluations director deems it necessary, the second intro process must begin within a week of the first process ending, and must last the same six (6) weeks.
```
mxmeinhold, line 526: `within a week of the first process ending` when is this? How does this factor in time for appeals?

Each sentence gets its own line.
```suggestion
The Introductory Process may be held twice a semester at the discretion of the evaluations director.
If the evaluations director deems it necessary, the second intro process must begin within a week of the first process ending, and must last the same six (6) weeks.
```

Maybe this could be wrapped into the earlier portion of this paragraph, it seems a little bit out of place, and the `same six (6) weeks` is weird.

mxmeinhold (changes requested):

mxmeinhold (dismissed):

constitution.tex
```diff
@@ -520,9 +520,11 @@
 \asubsubsection{The Introductory Process}
 The Introductory Process is designed to provide an easy means for Introductory Members to meet existing House members, learn House history, demonstrate their involvement potential to House, and allow existing House members to evaluate them for Active Membership.
 \asubsubsubsection{The Evaluation Period}
-The Process will begin between the first and fourth week of each semester, then last between six (6) and ten (10) weeks, starting when the Evaluations Director initiates it.
+The Process will occur either once or twice per semester, and will last for six (6) weeks.
```
Mstrodl, line 523: This allows for one intro process at the end of the semester... do we want that?

constitution.tex
```diff
@@ -520,9 +520,11 @@
...
+The Process shall be initiated by the Evaluations Director during either the first or second week of each semester.
```
galenguyer, line 524: With this wording, the introductory process can only begin for one week, either the first or second week of the semester. Some additional wording to allow packets to start for the first two weeks of the Introductory Process would be good.

constitution.tex
```diff
@@ -520,9 +520,11 @@
...
 The Executive Board may hold a closed ballot Simple Majority vote to extend the Introductory Process or the Introductory Packet.
 If an Introductory Process is extended, the Introductory Evaluation shall occur at the termination of the extended Process, which may be outside of the period defined in \ref{Introductory Evaluation}.
+If deemed necessary by the Evaluations Director, or by an Executive Board simple majority vote, the second intro process must begin within nine (9) days of the first process ending.
```
galenguyer, line 527: Some additional wording here to allow packets to start for the first two week of the second introductory process would be good

galenguyer (changes requested):

mxmeinhold:
I'm concerned that the timeline here may not allow for someone who fails to appeal and then reapply, given that eboard is normally after evals. Also, if evals continues to be on Monday, it leaves very little time to reapply in general.

crigney3:
> I'm concerned that the timeline here may not allow for someone who fails to appeal and then reapply, given that eboard is normally after evals. Also, if evals continues to be on Monday, it leaves very little time to reapply in general.

I believe there's nothing constitutionally to stop someone from reapplying and appealing at the same time - if the appeal of their evals decision succeeds, they would not need to continue the second process.

constitution.tex
```diff
@@ -520,9 +520,12 @@
 \asubsubsection{The Introductory Process}
 The Introductory Process is designed to provide an easy means for Introductory Members to meet existing House members, learn House history, demonstrate their involvement potential to House, and allow existing House members to evaluate them for Active Membership.
 \asubsubsubsection{The Evaluation Period}
-The Process will begin between the first and fourth week of each semester, then last between six (6) and ten (10) weeks, starting when the Evaluations Director initiates it.
+The Process will occur either once or twice per semester, and will last for six (6) weeks.
+The Process shall be initiated by the Evaluations Director during either the first and second week of each semester.
```
galenguyer, line 524: Either-And is really weird wording-wise

galenguyer (changes requested):

galenguyer:
> > I'm concerned that the timeline here may not allow for someone who fails to appeal and then reapply, given that eboard is normally after evals. Also, if evals continues to be on Monday, it leaves very little time to reapply in general.
> 
> I believe there's nothing constitutionally to stop someone from reapplying and appealing at the same time - if the appeal of their evals decision succeeds, they would not need to continue the second process.

There's nothing stopping them, but expecting someone to do packet while appealing seems.. less than ideal. If their appeal succeeds, then we asked them to do packet again for no reason.

geoffmoss0:
> I believe there's nothing constitutionally to stop someone from reapplying and appealing at the same time - if the appeal of their evals decision succeeds, they would not need to continue the second process

FWIW, I was definitely not in the right state to immediately restart the intro process a few days after I’d just failed, and I don’t think I’m alone in that

constitution.tex
galenguyer: Resolving this as the vote is now open and the PR cannot be changed. This can be fixed in a non-semantic change if the amendment passes.

galenguyer (approved):

galenguyer:
Amendment passed December 8th, 2021

PR #209, opened 2022-02-14
Add Active Member Requirement for Elections

jabbate19:
Check one:

Summary of change(s):

Requires Active Members to attend Elections with sufficient notice. Attendance will be taken in the same manner as a House Meeting, with a 5 day notice required by E-Board.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -795,7 +795,7 @@ The following special cases cover the operation of a dual directorship:
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
 	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
-		All Active Members are expected to attend candidate speeches as specified in \ref{Expectations of House Members}.
+		All Active Members are expected to attend candidate speeches in accordance with \ref{Expectations of House Members}.
 		If a member cannot attend due to a scheduling conflict, they may provide a reason for their absence to the Evaluations Director, who will record the reason with the absence.
 	\item Each candidate will be given an equal amount of time before the House to present their platform of candidacy.
 	\item Ballots will then be distributed and voting will be open for a minimum of a forty-eight hour period.
--- a/constitution.tex
+++ b/constitution.tex
@@ -796,7 +796,7 @@ The following special cases cover the operation of a dual directorship:
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
 	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
 		All Active Members are expected to attend candidate speeches as specified in \ref{Expectations of House Members}.
-		If a member cannot attend due to a scheduling conflict, they may provide a reason for their absence to the Evaluations Director, who will note is as an Absence with Reason.
+		If a member cannot attend due to a scheduling conflict, they may provide a reason for their absence to the Evaluations Director, who will record the reason with the absence.
 	\item Each candidate will be given an equal amount of time before the House to present their platform of candidacy.
 	\item Ballots will then be distributed and voting will be open for a minimum of a forty-eight hour period.
 		The Ballots will list, in random order, all of the candidates who are qualified for a given office with a means to indicate the selection of one of the candidates.
--- a/constitution.tex
+++ b/constitution.tex
@@ -794,7 +794,7 @@ The following special cases cover the operation of a dual directorship:
 	\item The candidates will be notified of their nomination.
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
-	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
+	\item The date and time of candidate speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
 		All Active Members are expected to attend candidate speeches in accordance with \ref{Expectations of House Members}.
 		If a member cannot attend due to a scheduling conflict, they may provide a reason for their absence to the Evaluations Director, who will record the reason with the absence.
 	\item Each candidate will be given an equal amount of time before the House to present their platform of candidacy.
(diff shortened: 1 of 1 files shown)


constitution.tex
```diff
@@ -794,6 +794,7 @@
 	\item The candidates will be notified of their nomination.
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
+	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. This will be treated as a House Meeting, with the expectation that all Active Members will attend and attendance will be taken. If the speeches take place at a traditional House Meeting, only one attendance will be taken.
```
mxmeinhold, line 797: I don't like `this will be treated as a House Meeting`. If you want to define this as a requirement, please actually do so, akin to the other requirements. You can keep a section here defining what qualifies as an election for required attendance.

mxmeinhold, line 797: Note: This increases the time between an election being announced and ballots being distributed from 4 days to 5 days.

constitution.tex
```diff
@@ -982,3 +983,4 @@
 The Judicial Board may present an official ruling following the Judicial Investigation.
 This ruling may be appealed in the same manner as an Executive Board decision \ref{Appeals}.
 \end{document}
+
```
mxmeinhold, line 986: nit: why is there an extra newline here?

constitution.tex
mxmeinhold, line 797: I'd also like to point out, house meetings are at an established time and are therefore announced (in effect) months in advance. 5 days isn't necessarily enough notice to clear a schedule for a saturday or sunday morning. This is part of why we generally do elections at house meetings, excepting the large scale elections at the end of the year, which generally have significant notice.

I'd also wonder, what's the goal of requiring attendance? Quorum isn't required for speeches, and this attendance requirement has minimal enforcement.

jabbate19: Active members should be involved in this process, as this is probably as important as anything that would be discussed at house meeting. Everyone should be making an effort to be there to make sure the best candidate is selected


Also if we need overrides at the speeches it would be much less of a rush, altho doing it as a form beforehand is likely an easier option

mxmeinhold: note: Overrides don't need to happen at speeches, they just need to happen before votes are counted, which means that most of the time they can occur at house meeting, when quorum is far easier to obtain.

I think whether or not this change goes through, the Chairperson docs should note that elections should be during house meetings where possible,

mxmeinhold, line 797: Also, please comply with the style conventions.

mxmeinhold (changes requested):

constitution.tex
```diff
@@ -794,6 +794,9 @@
 	\item The candidates will be notified of their nomination.
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
+	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
+		All Active Members are expected to attend this Election, as specified in \ref{Expectations of House Members}.
+		Exemptions to attendance may be given at the discretion of the Evaluations Director.
```
mxmeinhold, line 799: `Exemption` isn't well defined. `Exempt` as in "they qualify equivalently as having attended"? Or `Exempt` in the manner of house meeting excusals, which are simply "they were absent, but they can write down an excuse"?

mxmeinhold: Also nit: `s/given/granted`

constitution.tex
```diff
@@ -794,6 +794,9 @@
 	\item The candidates will be notified of their nomination.
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
+	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
+		All Active Members are expected to attend this Election, as specified in \ref{Expectations of House Members}.
```
mxmeinhold, line 798: I'd lean away from `Election`, it also isn't well defined. You're expecting members to attend speeches or `time before the House to present their platform of candidacy`, correct? I'd specify that, because Election is used differently elsewhere.

constitution.tex
```diff
@@ -795,8 +795,8 @@
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
 	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
-		All Active Members are expected to attend this Election, as specified in \ref{Expectations of House Members}.
-		Exemptions to attendance may be given at the discretion of the Evaluations Director.
+		All Active Members are expected to attend candidate speeches as specified in \ref{Expectations of House Members}.
+		If a member cannot attend due to a scheduling conflict, they may provide a reason for their absence to the Evaluations Director, who will note is as an Absence with Reason.
```
mxmeinhold, line 799: `note is`?

constitution.tex
```diff
@@ -794,6 +794,9 @@
 	\item The candidates will be notified of their nomination.
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
+	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
+		All Active Members are expected to attend candidate speeches as specified in \ref{Expectations of House Members}.
```
mxmeinhold, line 798: this reads as though candidate speeches are defined in expectations

jabbate19: Should I change `as specified in` to `in accordance of`?

geoffmoss0: `in accordance with` would be my choice

constitution.tex
```diff
@@ -794,6 +794,9 @@
 	\item The candidates will be notified of their nomination.
 		Each candidate is given a minimum of twenty-four hour period to accept or decline the nomination.
 		A list of all nominees who have accepted their nominations will be posted shortly thereafter.
+	\item The date and time of the candidate(s)'s speeches will be announced by a member of the Executive Board at least five (5) days prior to the event.
```
mxmeinhold, line 797: nit:
```suggestion
	\item The date and time of candidate speeches will be announced by a member of the Executive Board at least five (5) days prior to the event. 
```

mxmeinhold (approved):

SpencerCreveling:
Amendment Passed March 15th, 2022
Quorum - 42
Votes cast - 43
Pass - 33 votes (76.7%) (82.5% after abstentions)
Fail - 7 votes (16.3%) (17.5% after abstentions)
Abstain - 3 votes (7%)

PR #210, opened 2022-02-15
Allows Active Members to propose changes to the COC

SpencerCreveling:
Check one:

Summary of change(s):
Allows Active Members to propose changes to the COC

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -738,7 +738,7 @@ Members may sign a more recent revision of the Code of Conduct to update their a
 \asubsubsubsection{Changes}
 The following process is used for making changes to the Code of Conduct document.
 \begin{enumerate}
-	\item A Root Type Person drafts a change to the Code of Conduct and makes publicly available both the summary and difference file of the change.
+	\item An Active Member drafts a change to the Code of Conduct and makes publicly available both the summary and difference file of the change.
 	\item The change is proposed at a House Meeting and is discussed at the same House Meeting.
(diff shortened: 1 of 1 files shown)


constitution.tex
```diff
@@ -738,7 +738,7 @@
 \asubsubsubsection{Changes}
 The following process is used for making changes to the Code of Conduct document.
 \begin{enumerate}
-	\item A Root Type Person drafts a change to the Code of Conduct and makes publicly available both the summary and difference file of the change.
+	\item An Active Member drafts a change to the Code of Conduct and makes publicly available both the summary and difference file of the change.
```
mxmeinhold, line 741:
```suggestion
	\item An Active Member or Root Type Person drafts a change to the Code of Conduct and makes publicly available both the summary and difference file of the change.
```

SpencerCreveling: I see where you are coming from as a non-active RTP, however I feel that matters regarding the COC are primarily the concern of active members. Once an RTP leaves Active Member status they shouldn't concern themselves with the COC.

mxmeinhold: This suggests that RTPs who were not active members have not been involved in day to day matters, which is simply not the case in recent memory.

geoffmoss0: Reminder that not active != not at RIT

jabbate19: IIRC Non-Active RTPs can't vote on Constitutional Amendments, and modifying the COC is of equal stature as the Constitution.

mxmeinhold: Also, RTPs do not need to be active when they become RTPs

PR #213, opened 2022-02-15
fixes the wording to eliminate edge case in 5.D.3.C.b

SpencerCreveling:
Check one:

Summary of change(s):
fixes the wording to eliminate edge case that allows active members who have not passes an intro evals to be a required signature

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -527,7 +527,7 @@ If an Introductory Process is extended, the Introductory Evaluation shall occur 
 If deemed necessary by the Evaluations Director, or by an Executive Board simple majority vote, the second intro process must begin within nine (9) days of the first process ending.
 If a second Introductory Process is approved, the Evaluations Director may initiate a new Introductory Packet within the first or second week of the Introductory Process.
 \asubsubsubsection{The Introductory Packet}
-Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain a signature from all Active members, all Introductory Resident members, and ten (10) of any combination of the following:
+Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain a signature from all Active Members who have passed a Membership Evaluation, all Active Resident Members, all Introductory Resident Members, and ten (10) of any combination of the following:
 \begin{itemize}
 	\item Alumni members
 	\item Honorary members


galenguyer:
On-floor freshman during the first semester are Resident members but not Active members who have not passed a Membership Evaluations

mxmeinhold (commented):
How is this an edge case? The present wording is current practice.

SpencerCreveling:
because of section 5.D.2.A.4.B.a an intro member that passes into evals is considered an active member, if that active member is off floor and thus a non-Resident member then they are still technically a required signature according to the current wording.

mxmeinhold:
You should amend 5.D.2.A.2 to align with current practice instead.

SpencerCreveling:
their is no need as 5.D.2.A.2 perfectly reflects the current way we set up packet. the issue is not who can an cannot signe a packet the issue is who does and doesn't count towards the bare minimum 60% for account creation. which is covered in 5.D.3.C.b

constitution.tex
```diff
@@ -527,7 +527,7 @@
 If deemed necessary by the Evaluations Director, or by an Executive Board simple majority vote, the second intro process must begin within nine (9) days of the first process ending.
 If a second Introductory Process is approved, the Evaluations Director may initiate a new Introductory Packet within the first or second week of the Introductory Process.
 \asubsubsubsection{The Introductory Packet}
-Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain a signature from all Active members, all Introductory Resident members, and ten (10) of any combination of the following:
+Each Introductory Process participant, after the first week, is given two (2) weeks to return a document that should contain a signature from all Active Members who have passed a Membership Evaluation, all Active Resident Members, all Introductory Resident Members, and ten (10) of any combination of the following:
```
mxmeinhold, line 530: nit: might be better to say "Resident Active" etc, but otherwise I think you've cleaned up the holes in your previous commit.

galenguyer (approved):

SpencerCreveling:
Amendment #213, fixes the wording to eliminate edge case in 5.D.3.C.b, has passed with the following results!
Quorum - 42 votes (2/3rds of 64)
Votes cast - 46
Threshold to pass - 29 votes (2/3rds of 44 non-abstain votes cast)
Pass - 44 votes (95.7%) (100% after abstentions)
Fail - 0 votes (0%)
Abstain - 2 votes (4.3%)

PR #224, opened 2022-08-27
Removes ability for RTP's to override requirements

SpencerCreveling:
Check one:

Summary of change(s):
removing the ability for RTP's to override the requirements outlined in the constitution to elect a new RTP. Would require a constitutional override instead if an ineligible RTP was to be elected.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -724,8 +724,6 @@ Membership is composed of all Root Type Persons.
 \renewcommand{\theenumi}{\alph{enumi}} % For this section, we want items to use letters
 \begin{enumerate}
 	\item Candidates must have been an On-Floor Active Member in good standing within the last twelve months, or been an Off-Floor Active Member and have lived on the floor within the last twelve months.
-	\item Candidates may be granted an exemption by current Root Type Persons.
-		Such an exemption may be withdrawn at any time by the current Root Type Persons.
 	\item Prior Root Type Persons are those members who are no longer current Active Members and have not been granted an extension by the current Root Type Persons.
 		Prior Root Type Persons are not guaranteed access to the current root passwords and other authentication tokens.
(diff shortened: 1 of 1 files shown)


geoffmoss0:
Wasn’t this going to be in conjunction with removing requirements for Opcomm? Should those two changes be part of the same amendment?

SpencerCreveling:
I'm not really sure what your referring to Geo, but usually I don't like grouping amendments together because some members might want to pass one part of the amendment and not the other.

mcsaucy:
What problem does this PR solve?

Also would this apply retroactively?

Folks who fall out of good standing (or graduate) would fail the meet the requirements of RTP candidacy, which probably isn't ideal given 1) de-RTPing requires a toil-heavy credential rotation; 2) retaining volunteer sysadmins has historically been hard enough as is; and 3) y'all have way more systems to maintain than you did a year ago which suggests the need for more bodies, not fewer. 

Seems like there's value in making it easier to give folks root so they can help out, so why add process?

constitution.tex
```diff
@@ -724,8 +724,6 @@
 \renewcommand{\theenumi}{\alph{enumi}} % For this section, we want items to use letters
 \begin{enumerate}
 	\item Candidates must have been an On-Floor Active Member in good standing within the last twelve months, or been an Off-Floor Active Member and have lived on the floor within the last twelve months.
```
galenguyer, line 726: Spencer and I talked a few weeks ago and agreed that removing the RTP's ability to waive requirements would be reasonable if we adjusted the requirements to also be more reasonable. I suggested and we agreed upon the following change:
```suggestion
	\item Candidates must have been an Active Member within the last twelve months.
```

galenguyer (changes requested):

esoccoli:
Resolved by #232

PR #227, opened 2023-04-19
Removed line stating that failed intro members had to move off floor in section 5.D.2.A.4.C

esoccoli:
…in section 5.D.2.A.4.C

Check one:

Summary of change(s):
Removed line that says that intro members that fail intro evals need to move off floor so that it matches current practice


esoccoli:
Resolved by #232

PR #237, opened 2024-09-16
prevent multiple simultaneous memberships

Qelxiros:
Check one:

Summary of change(s):

This change prevents current active and alumni members from reentering the introductory process, thereby preventing an opportunity for constitutional ambiguity.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -151,7 +151,7 @@ When describing membership types, the following terms are used:
 \asection{Introductory Membership}
 
 \asubsection{Introductory Membership Qualifications}
-Intro Membership is open to all students currently enrolled at RIT.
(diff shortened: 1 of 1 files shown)


rhochgraf21 (commented):
This handles the cases I was anticipating. One question however: should Advisory members be allowed to become intro members if they become students by taking a class alongside their work duties at RIT?

I would not expect that RIT would permit this, but do we want to leave the door open to that from our end?

ikopke23:
I would argue that since Advisory Members are not formally excluded in the statement they could become intro members.

Qelxiros:
Yep, I don't think we should prohibit Advisory Members from entering the intro process because it doesn't create loopholes in the same way as Alumni/Active Members. It seems unlikely that RIT would allow it, but it's not the place of this document to codify that.

esoccoli (approved):
lgtm

PR #244, opened 2024-10-03
clarify voting members

Qelxiros:
Check one:

Summary of change(s):
Change the current Expectations of Voting Members section to be the Qualifications section, and add the expectation that voting members vote at six weeks.
Also add a clause that allows voting members to opt out of six weeks, thus lowering quorum.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -254,7 +254,7 @@ If the vote passes, the Alumni is reinstated as an Active Member with Off-Floor 
 
 \asubsection{Active Membership Expectations}
 Active Members are expected to be active participants in CSH as defined in \ref{Expectations of Active Members}.
-Active Voting Members are expected to meet the requirements defined in \ref{Expectations of Voting Members}.
+Active Voting Members are eligible to become Voting Members according to the requirements defined in \ref{Qualifications of Voting Members}.
 
 \asubsubsection{Expectations of Active Members}
 Active Members are required to pay dues as stated in \ref{Collection of Dues}, attend all House Meetings and E-Board candidate speeches, and attend at least fifteen of the directorship meetings for each semester in which they are an Active Member.
@@ -275,10 +275,10 @@ Active Members receive the privilege to:
 	\item Receive priority for on-floor housing, or apply for On-Floor Status
 \end{enumerate}
 
-\asubsubsection{Expectations of Voting Members}
+\asubsubsection{Qualifications of Voting Members}
 \renewcommand{\theenumi}{\arabic{enumi}} % For this section, we want items to use letters
 
-The following requirements must be met by the beginning of the Intro Eval for an Active Member to be allowed to vote.
+For an Active Member to be allowed to vote in an Intro Eval, they must meet the following requirements by the beginning of the Intro Eval.
 Any of these requirements may be waived by the Evals Director or an E-Board Vote.
 \begin{enumerate}
 	\item Attend all House Meetings during the Intro Process
@@ -286,6 +286,9 @@ Any of these requirements may be waived by the Evals Director or an E-Board Vote
 	\item Attend at least one CSH social event during the Intro Process
 	\item Attend at least two Technical Seminars during the Intro Process
 \end{enumerate}
+
+\asubsubsection{Expectations of Voting Members}
+Voting Members are expected to attend and vote at the Intro Eval for which they have qualified. If they cannot attend the Intro Eval, they are expected to notify the Evals Director before the Intro Eval.
 
 \asubsection{Active Membership Evaluations}
 Active Members are evaluated annually through the Eval Process described below.


constitution.tex
```diff
@@ -254,7 +254,7 @@ \subsection{Modifying the Constitution}
 
 \asubsection{Active Membership Expectations}
 Active Members are expected to be active participants in CSH as defined in \ref{Expectations of Active Members}.
-Active Voting Members are expected to meet the requirements defined in \ref{Expectations of Voting Members}.
+Active Voting Members are eligible to become Voting Members according to the requirements defined in \ref{Qualifications of Voting Members}.
```
Mstrodl, line 257: As written, this suggests that you are not a voting member until you meet these requirements

Qelxiros: That's correct. If you don't meet these requirements, you can't vote at six weeks.

Mstrodl: I got confused about "Voting Members" referring to members who are active and can vote on house issues, but I get it now

rhochgraf21, line 257: How about:

Active Members are eligible to become Active Voting Members if they meet the requirements defined in \ref{Qualifications of Voting Members}.

rhochgraf21: (This change ensures we use only existing definitions.)

constitution.tex
```diff
@@ -275,18 +275,23 @@ \subsection{Modifying the Constitution}
 	\item Receive priority for on-floor housing, or apply for On-Floor Status
 \end{enumerate}
 
-\asubsubsection{Expectations of Voting Members}
+\asubsubsection{Qualifications of Voting Members}
 \renewcommand{\theenumi}{\arabic{enumi}} % For this section, we want items to use letters
 
-The following requirements must be met by the beginning of the Intro Eval for an Active Member to be allowed to vote.
-Any of these requirements may be waived by the Evals Director or an E-Board Vote.
+For an Active Member to be allowed to vote in an Intro Eval, they must meet the following requirements by the beginning of the Intro Eval.
```
rhochgraf21, line 281: Is the change to make Active Members who fail gatekeep keep voting rights in non-intro-eval settings intentional?

constitution.tex
```diff
@@ -275,18 +275,23 @@ \subsection{Modifying the Constitution}
...
+Any of these requirements may be waived by the Evals Director or an E-Board Vote. An Active Member who meets these requirements may opt out of Voting Membership by submitting a request to the Evals Director or E-Board.
```
rhochgraf21, line 282: Formatting - style recommendation to place individual sentences on separate lines.

constitution.tex
rhochgraf21: If the intended result is to make the change noted above, then perhaps it would be good to specify that the only voting rights being modified are those at the Intro Eval. That would also change line 257 to be something like:

Active Members are eligible to vote at an Intro Eval if they meet the requirements defined in \ref{Qualifications of Voting Members}.

constitution.tex
```diff
@@ -275,18 +275,23 @@ \subsection{Modifying the Constitution}
...
 \begin{enumerate}
 	\item Attend all House Meetings during the Intro Process
 	\item Attend at least one directorship meeting for each week of the Intro Process % The number of directorship meetings intended here is equal to the number of the weeks in the process, not just one each week
 	\item Attend at least one CSH social event during the Intro Process
 	\item Attend at least two Technical Seminars during the Intro Process
 \end{enumerate}
 
+\asubsubsection{Expectations of Voting Members}
+\begin{enumerate}
+	\item Attend and vote at the Intro Eval for which they have qualified.
```
rhochgraf21, line 292: formatting rec - merge into above list.

rhochgraf21 (changes requested):
Couple of questions, couple of proposed language changes, couple of formatting things to be fixed.

constitution.tex
Qelxiros: These are two distinct lists with distinct meanings.

constitution.tex
Qelxiros: I don't know what you want specifically. Currently, 4.B.4.A is consistent with 4.B.3.A and 4.B.5.A. If you suggest changes, I'll review them (or just add a commit).

constitution.tex
Qelxiros: It is intended, but I don't think there's currently any ambiguity. We could call them Introductory Voting Members instead if you think that's clearer.

constitution.tex
rhochgraf21: Line 282 has two sentences on it; the style guidelines would recommend one sentence per line.

constitution.tex
rhochgraf21: I think they can be merged easily, and then we don't the very strange case where we have an enumerated list with one item.

For example, instead of 290-292, what if we put "Attend and vote at the Intro Eval, if eligible." at line 288?

constitution.tex
```diff
@@ -254,7 +254,7 @@ \subsection{Modifying the Constitution}
 
 \asubsection{Active Membership Expectations}
 Active Members are expected to be active participants in CSH as defined in \ref{Expectations of Active Members}.
-Active Voting Members are expected to meet the requirements defined in \ref{Expectations of Voting Members}.
+Active Voting Members are eligible to become Voting Members if they meet the requirements defined in \ref{Qualifications of Voting Members}.
```
rhochgraf21, line 257: This language is still confusing to me, since Voting Members is not defined (or used in this phrasing) elsewhere.

How about something like:

Active Members are eligible to vote at an Intro Eval if they meet the requirements defined in \ref{Qualifications of Voting Members}.

rhochgraf21 (changes requested):

constitution.tex
rhochgraf21: Also, please see above unresolved threads.

constitution.tex
Qelxiros: The first list is Qualifications; the second is expectations. Merging them doesn't make sense to me. I also don't see an issue with having a one-element list.

constitution.tex
Qelxiros: Voting Membership is defined in the section I'm referencing here, but I've removed the reference to Active Voting Membership.

Qelxiros: Specifically, what I've written is different from what you're suggesting in that voting at six weeks is an expectation of all who are eligible. That's kinda the point of this amendment, so I'll resolve this, but feel free to reopen it if you have further thoughts.

PR #246, opened 2024-10-04
prevent the number of eligible votes from changing during a vote

rhochgraf21:
Check one:

Summary of change(s):

Fixes an edge case where the number of people eligible to vote on an issue _might?_ be able to change during the execution of the vote.

Does not fix the issue raised by @jabbate19. My interpretation is that the Intro Evaluation is a vote to extend an _offer_ of active membership, as opposed to an immediate granting of active membership, so I don't see an issue there. I wouldn't be against an additional amendment to make that part more clear, however.

Please note that with these changes, quorum does not change during the duration of a vote either, since it is a fraction of the number of eligible votes.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -125,12 +125,17 @@ A Maintainer may approve any proposed change that does not affect the meaning of
 Alternatively, the change may be presented at a House Meeting for discussion followed by an Immediate One-Half Vote with fifty percent quorum.
 
 \asubsection{Semantic Changes}
+\asubsubsection{Amendments}
 Any semantic change to the Constitution requires the change to be proposed in writing for discussion at a House Meeting.
 Any modifications made due to the discussion are added to the written proposal and the modified proposal is posted during the week.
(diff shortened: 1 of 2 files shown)


Qelxiros (changes requested):
This should probably say that neither the number of eligible votes nor the set of eligible voters can change during a vote.

rhochgraf21:
updated to reflect comments.

constitution.tex
```diff
@@ -900,6 +900,7 @@ \subsection{Modifying the Constitution}
 
 \asubsection{Eligible Votes}
 The number of Active Members eligible to vote on the issue.
+The eligible voters for a vote are determined and fixed when that vote is called.
```
mxmeinhold, line 903: Hmm, so if someone becomes active during the ballot period for an amendment, they would not be eligible? Or how is called defined here? Called for or called as resolved?

rhochgraf21: good point. changed wording.

shaeespring:
Voting Members: 69
2/3 quorum: 46
2/3 to pass: 32/48

Got: 46/48

Passed ✅

shaeespring (approved):
Clarifies quorum determining at the beginning of the vote for all votes

PR #247, opened 2024-10-07
add explicit mechanisms for housing priority / access

rhochgraf21:
Check one:

Summary of change(s):

NTID students can only select dorms with doorbells, so this amendment gives tiebreaking priority to NTID students to help ensure they have equal access to rooms in CSH.

The amendment adds additional explicit mechanisms for housing priority to ensure equal access to housing. To the best of my knowledge, these are mostly in accordance with recent practice, with the exception of making them additionally available at the discretion of the Evaluations director.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -881,10 +881,11 @@ They still have access to all other privileges associated with their membership,
 The Housing Priority System is a means for determining the priority a member has in a housing issue or the assignment of available housing.
 The member with top priority is the member with On-Floor Status and the most Housing Priority Points.
 Housing Priority Points are accumulated once per Operating Session at the conclusion of the Membership Eval.
-Each Active Member who passes the Membership Eval is granted two Housing Priority Points.
+Each Active Member who passes the Membership Eval is granted two Housing Priority Points. 
 Members registered with NTID receive priority over members with equivalent Housing Priority points. % NTID students can ONLY choose dorms with doorbells, so this helps ensure that students who need these rooms have access to them. 
 Members who require accommodations may be temporarily granted additional Housing Priority Points at the request of RIT Housing or at the discretion of the Evaluations Director. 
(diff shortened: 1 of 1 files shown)


charlottegeo:
Should we add NTID to the list of abbreviations?

constitution.tex
```diff
@@ -881,7 +881,11 @@ \subsection{Modifying the Constitution}
 The Housing Priority System is a means for determining the priority a member has in a housing issue or the assignment of available housing.
 The member with top priority is the member with On-Floor Status and the most Housing Priority Points.
 Housing Priority Points are accumulated once per Operating Session at the conclusion of the Membership Eval.
-Each Active Member who passes the Membership Eval is granted two Housing Priority Points.
+Each Active Member who passes the Membership Eval is granted two Housing Priority Points. 
+Members registered with NTID receive priority over members with equivalent Housing Priority points. % NTID students can ONLY choose dorms with doorbells, so this helps ensure that students who need these rooms have access to them. 
+Members who require accommodations may be temporarily granted additional Housing Priority Points at the request of RIT Housing or at the discretion of the Evaluations Director. 
+Temporary Housing Priority Points remain in effect until the end of the Standard Operating Session or an earlier date, as specified when granted.
```
shaeespring, line 887: ` an earlier date, as specified when granted.`

Is there a case in which this happens? This PR is a bit on the older side so it might just not be relevant anymore but I also might be crazy

shaeespring:
> Should we add NTID to the list of abbreviations?

Nah RIT isn't in the list

constitution.tex
shaeespring: TECHNICALLY Housing points used to determine rooms that OPEN UP on floor during the school year should also have people that are granted temporary housing points, but in that case, would evals remove those temporary housing points before the end of the SOS?

shaeespring: That's a question for @Qelxiros @ikopke23 and @mcm6954

shaeespring (approved):

PR #260, opened 2025-05-21
voting member requirements

shaeespring:
- **voting member requirements**

Check one:

Summary of change(s):

Eligible Votes contains a reference to Voting Member Requirements (gatekeep)
The wording of Section 5.B.4 removes the Eligible Votes, and includes Voting Eboard Members

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -899,7 +899,7 @@ This section outlines the different types of votes and ballots used to make deci
 
 \asubsection{Eligible Votes}
 The number of Active Members who are eligible to vote on the issue.
-Following the first Intro Eval of the academic year, Active Members are only eligible to vote if they have completed the qualifications listed in \ref{Expectations of Voting Members}.
+Following the first Intro Eval of the academic semester, Active Members are only eligible to vote if they have completed the qualifications listed in \ref{Expectations of Voting Members}.
 Officers of CSH retain voting eligibility on all matters directly pertaining to the duties of their held office regardless of \ref{Expectations of Voting Members}.
 
 \asubsection{Votes Cast}
--- a/constitution.tex
+++ b/constitution.tex
@@ -280,13 +280,13 @@ Active Members receive the privilege to:
 \asubsubsection{Expectations of Voting Members}
 \renewcommand{\theenumi}{\arabic{enumi}} % For this section, we want items to use letters
 
-The following requirements must be met by the beginning of the Intro Eval for an Active Member to be allowed to vote.
+The following requirements must be met for an Active Member to be allowed to vote after the first Intro Evals of the academic year.
 Any of these requirements may be waived by the Evals Director or an E-Board Vote.
 \begin{enumerate}
-	\item Attend all House Meetings during the Intro Process
-	\item Attend at least one directorship meeting for each week of the Intro Process % The number of directorship meetings intended here is equal to the number of the weeks in the process, not just one each week
-	\item Attend at least one CSH social event during the Intro Process
-	\item Attend at least two Technical Seminars during the Intro Process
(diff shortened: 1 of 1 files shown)


constitution.tex
```diff
@@ -898,7 +898,7 @@ \subsection{Modifying the Constitution}
 \asection{Definitions}
 
 \asubsection{Eligible Votes}
-The number of Active Members eligible to vote on the issue.
+The number of Active Members who are eligible to vote on the issue, and after the Intro Eval, have completed the qualifications listed in \ref{Expectations of Voting Members}.
```
rhochgraf21, line 901: This particular phrasing strikes me as complex. Maybe we can split it in two for readability? Perhaps something like...

```
The number of Active Members who are eligible to vote on the issue.
Following the first Intro Eval of an academic year, Active Members are only eligible to vote if they have completed the qualifications listed in \ref{Expectations of Voting Members}.
```

perhaps? 

--

Other thoughts:
- there's an unfortunate edge case for E-Board members possibly not being able to vote on E-Board issues, it seems. (double-check me on this, but this doesn't seem apply to RTPs due to the way quorum is specified there.)
- potential fix for this edge case:  ```7.B.2.D [...] Unless otherwise specified, E-Board votes are Immediate Relative Majority Votes with a fifty percent quorum [+ of Voting E-Board Members.]```. That said, this feels like a patch rather than fixing the underlying issue.
- proposal for fixing underlying issue: ``` [+Officers of CSH retain voting eligibility on all matters directly pertaining to the duties of their held office regardless of \ref{Expectations of Voting Members}.] ```

rhochgraf21 (changes requested):
a few thoughts & wording proposals

constitution.tex
rhochgraf21: Oh, this might be something to consider.

4.B.4.A Expectations of Voting Members is phrased as all "during the Intro Process." 

So if a member does not complete the requirements during the Intro Process, but completes them later, they do not regain the right to vote.

shaeespring: I'd rather use the last recommendation to fix the issue. It makes the most sense.

shaeespring: I think it should be discussed whether we want to have eligible votes reset at the beginning of the second semester (this is terrible wording, but to proffer: do we want the requirements to need fulfilled again for the second semester? That was my original thought, but I think your suggestion of the first Intro Eval of the academic year may be sufficient.

Qelxiros: If we have it reset, we should make sure we're correctly handling the case where we have a second Intro Eval process in a given semester.

If we don't have it reset, I think there should be a grace period (the first six weeks of the second semester) during which active members who were inactive during the previous semester (co-op or otherwise) can vote without meeting requirements. That might be controversial, and there will almost certainly be contention around the exact implementation of that sentiment, so I'll leave that up to you.

constitution.tex
```diff
@@ -741,7 +741,7 @@ \subsection{Modifying the Constitution}
 
 \asubsection{Constitutional Maintainer Selection}
 Any member may nominate a qualified member for Maintainer status to E-Board for consideration.
-E-Board may choose to approve or reject the nomination by E-Board Vote with a quorum of seventy-five percent of the Eligible Votes.
+E-Board may choose to approve or reject the nomination by E-Board Vote with a quorum of seventy-five percent of Voting E-Board Members.
```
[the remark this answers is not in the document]
shaeespring: I think I want to focus this change on Eligible Votes, and I don't want to fix the strangeness of the 75% quorum with a completely different subject. I'd rather make a separate amendment to fix that issue. But your first proposal removes redundancy, and I like that

PR #243, opened 2024-10-03
remove the social event requirement from gatekeep

Qelxiros:
Check one:
- [x] Semantic Change: something about the meaning of the text is different
- [ ] Non-semantic Change: Spelling, grammar, or formatting changes.

Summary of change(s):
Remove the social event requirement for upperclassmen. This isn't tracked, so this change would bring the constitution up to date with current practice.

PR #264, opened 2025-10-23
Remove Social Event Requirements

DeadPixelPro:
Check one:

Summary of change(s):
Removed the CSH Social Event requirement from both Intro Member and Voting Member expectations.

- Social event attendance is not something we track
- If you don't attend a single social event as an intro member, you're unlikely to pass anyways
- It can be confusing to an intro member as to what counts as a "Social Event"

This differs from https://github.com/ComputerScienceHouse/Constitution/pull/243/files (which failed to pass), as it removes the social event requirement from both intro and voting expectations. I believe the fact that it held upperclassmen to a different standard to intro members was the primary reason this failed.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -212,7 +212,6 @@ Before the end of the Intro Process, an Intro Member is expected to:
 	\item Attend all House Meetings during the Intro Process
 	\item Complete the Intro Packet
 	\item Attend at least one directorship meeting for each week of the process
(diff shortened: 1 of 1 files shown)


shaeespring:
Screenshot_20251106_101549
Active Voting Members: 71
Quorum:48
Needed to pass (out of 56): 38
Has: 39 Pass
Vote Passes

PR #270, opened 2025-11-22
dual chair, evals, and financials not allowed

shaeespring:
Check one:

Summary of change(s):
Dual Chair allows for a vote to be tied if E-Board ties and also both chairs disagree
Dual Secretary (the only other non-voting E-Board position) doesn't need to be dual. Needs discussion

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -592,7 +592,7 @@ However, the Chair and at least two-thirds of the Voting Members of E-Board must
 	\item CSH may choose to waive the following subset of the qualifications for E-Board positions, as defined under \ref{Qualifications}, for all candidates:
 	      \begin{enumerate}
 		      \item Candidates must reside on floor during the term of office.
-		      \item Social and R\&D are the only voting E-Board positions that allow for dual directorship.
+		      \item Social and R\&D are the only E-Board positions that allow for dual directorship.
 	      \end{enumerate}
 	\item When a waiver is proposed, the Chair of the Vote shall state which E-Board position and qualification is being voted upon, and any nominated candidates not qualified under \ref{Qualifications} who would become qualified if the waiver passes.
 	\item A Three-Quarters Immediate Vote with three-quarters quorum is then taken to determine whether the proposed waiver shall take effect.
@@ -603,7 +603,7 @@ However, the Chair and at least two-thirds of the Voting Members of E-Board must
 \asubsection{Selection}
 
 \asubsubsection{Dual Directors}
-Social and R\&D are the only Voting E-Board positions that allow for dual directors.
+Social and R\&D are the only E-Board positions that allow for dual directors.
 
 If two candidates elect to run as dual directors, their names are placed together on a single line of the election ballot.
 If they are also nominated as a single directorship or as dual directors with another member, their votes are not cumulative.
--- a/constitution.tex
+++ b/constitution.tex
@@ -598,13 +598,13 @@ However, the Chair and at least two-thirds of the Voting Members of E-Board must
 	\item A Three-Quarters Immediate Vote with three-quarters quorum is then taken to determine whether the proposed waiver shall take effect.
 	\item Each vote to waive a qualification must only apply to a single qualification for a single E-Board position.
 	\item Waivers always apply to all candidates for the E-Board position.
+	\item The waiver for dual directorship may never be applied to the positions of Chair, Evals, or Financials.
 \end{enumerate}
 
 \asubsection{Selection}
 
 \asubsubsection{Dual Directors}
-Social and R\&D are the only E-Board positions that allow for dual directors.
-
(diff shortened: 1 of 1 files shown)


ziti-zd:
Has their ever been a dual Chair? Additionally, how would that even work with the Chair now acting as the facilitator of dialogue at House Meetings (the selecting of/declaring the changing of topics that came with the clarifying question changes)?

I am in favor of these changes because I do not see any harm in them but want to better understand the context behind the loophole it is trying to remove.

shaeespring:
No, there has never been a dual chair (Edit: since 1981. I'm not sure how to check before that) but there have been attempts. (One as recent as last year)

A dual chair wouldn't work. For the reason you brought up, but also because the chair is the tiebreaker in E-Board votes in the event of a tie. If there are two chairs, then you can have a tie of an E-Board vote.

In order for it to work, you would need to designate roles to each Chairperson, such as "this person can vote, this person is in charge of discussion, this person is point of contact for E-Board" etc. 

We're also currently trying to talk to our advisor to find out if a Dual Chair is even allowed, as RIT policy might restrict it. In this case, all of Chair, Evals and Financials might be restricted to single directors.

constitution.tex
```diff
@@ -592,19 +592,19 @@ \subsection{Modifying the Constitution}
...
 	\item Waivers always apply to all candidates for the E-Board position.
+	\item The waiver for dual directorship may never be applied to the positions of Chair, Evals, or Financials.
 \end{enumerate}
 
 \asubsection{Selection}
 
 \asubsubsection{Dual Directors}
-Social and R\&D are the only Voting E-Board positions that allow for dual directors.
-
+Social and R\&D are the only E-Board positions that allow for dual directors. The positions of Chair, Evals, and Financials must never be held by dual directors.
```
```suggestion
Social and R\&D are the only E-Board positions that allow for dual directors. The positions of Chairperson, Evals Director, and Financial Director must never be held by dual directors.
```

shaeespring: Or just change Financials to Financial, but that sounds weird

shaeespring: These are suggestions. I don't grammar well so I might just be wrong. Socials is Social in every place except one (Pr to come in the next couple of hours) so I believe Financials should be consistent with that. IMPs is IMPs though

DeadPixelPro: If you're adding Evals and Financials to this, I'd suggest changing the title of the PR to reflect.
(Unless I'm reading the diff wrong)

charlottegeo: @DeadPixelPro you read the diff right, Evals and Financials cannot be dual according to RIT.

charlottegeo: and for the grammar I am not sure how to phrase the directorship stuff either, more input would be great

charlottegeo: changed the title

shaeespring: @Qelxiros @ikopke23

constitution.tex
```diff
@@ -592,19 +592,19 @@ \subsection{Modifying the Constitution}
 	\item CSH may choose to waive the following subset of the qualifications for E-Board positions, as defined under \ref{Qualifications}, for all candidates:
 	      \begin{enumerate}
 		      \item Candidates must reside on floor during the term of office.
-		      \item Social and R\&D are the only voting E-Board positions that allow for dual directorship.
+		      \item Social and R\&D are the only E-Board positions that allow for dual directorship.
```
shaeespring: This would also prohibit Adhocs from running dual now that I'm realizing it

constitution.tex
shaeespring: From Jeremy: This line is redundant because it's already prohibited from line 601

ikopke23:
I feel that explicitly saying that those directorships cannot be dual feels useless when we state `Socials and R&D are the only directorships allowed` but if you want to be more explicit there's nothing wrong with it. We should use abbreviations because we have them to Shaela's suggestion however it does read better that way

shaeespring (commented):
thank you for edits

constitution.tex
ikopke23: I didn't even see 601, remove the last change 601 is good

pikachu0542:
Have we gotten official word that dual chair/evals/financials would not be allowed by RIT? If not, I personally don't really see a need for this, at least for Evals and Financials. I believe we still have to fix a loophole that allows dual chair without a waiver, but aside from that, idk if this is necessary.

shaeespring:
> Have we gotten official word that dual chair/evals/financials would not be allowed by RIT? If not, I personally don't really see a need for this, at least for Evals and Financials. I believe we still have to fix a loophole that allows dual chair without a waiver, but aside from that, idk if this is necessary.

Yes, we have gotten word from Paul. See slack

pikachu0542 (approved):
lgtm

DeadPixelPro:
image

DeadPixelPro:
image

PR #287, opened 2026-03-25
Non-semantic changes while making my other PR

tallen42:
Check one:

Summary of change(s):

I fixed a few wordings, removed a comma, and added a period.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -299,35 +299,35 @@ Active Members are evaluated annually through the Membership Evaluation process 
 
 % ACTIVE MEMBERSHIP EVALUATIONS
 \asubsubsection{Membership Evaluation}
-The Membership Evaluation process ("Membership Eval") occurs once per academic year, where Active Members are evaluated.
-The Membership Eval takes place during the spring semester to comply with the RIT Housing deadlines.
-The Membership Eval is responsible for determining individuals who will have the option to continue Active Membership in the following academic year.
+The Eval Process occurs once per academic year, where Active Members are evaluated.
+The Eval Process takes place during the spring semester to comply with the RIT Housing deadlines.
+The Eval Process is responsible for determining individuals who will have the option to continue Active Membership in the following academic year.
 This meeting is open only to Active Members.
 All Active Members are expected to attend this meeting.
 
 \asubsubsubsection{Membership Evaluation Voting}
-Any Active Member who has completed all the requirements as defined in \ref{Expectations of Active Members} at the beginning of the Membership Eval passes their Membership Evaluation without being voted on.
-All Active Members who have not received an exemption from E-Board prior to the Membership Eval will be evaluated on a member-by-member basis by Active Members.
-The Membership Eval shall hold members to the objective requirements defined in \ref{Expectations of Active Members} and determine which members may continue as an Active Member in the following academic year.
+Any Active Member who has completed all the requirements as defined in \ref{Expectations of Active Members} at the beginning of the Eval Process pass the Eval Process without being voted on.
+All Active Members who have not received an exemption from E-Board prior to the start of the Eval Process will be evaluated on a member-by-member basis by Active Members.
+The Eval Process shall hold members to the objective requirements defined in \ref{Expectations of Active Members} and determine which members may continue as an Active Member in the following academic year.
 Exceptions to the requirements may be made, as attending members may choose any of the outcomes for each member, even if the member being evaluated has not completed all of their requirements.
-These requirements must be completed before the Membership Eval occurs.
+These requirements must be completed before the Eval Process occurs.
 A Secret Immediate Relative Majority vote with two-thirds quorum is required for the evaluation to be valid.
 Neither absentee nor proxy votes are allowed.
 
 \asubsubsubsection{Membership Evaluation Outcomes}
-Members who pass the Membership Eval have the option to participate as an Active Member in the following year.
+Members who pass the Eval Process have the option to participate as an Active Member in the following year.
 \\* \\*
-If a member fails and has never passed a Membership Eval in the past, their membership will be revoked upon completion of the Membership Eval.
-If the member fails, but has previously passed a Membership Eval, they will move to Alumni Membership in bad standing, as defined in \ref{Alumni Membership Selection}, at the end of the Standard Operating Session.
+If a member fails and has never passed an Eval Process in the past, their membership will be revoked upon completion of the Eval Process.
+If the member fails, but has previously passed an Eval Process, they will move to Alumni Membership in bad standing, as defined in \ref{Alumni Membership Selection}, at the end of the Standard Operating Session.
 In either case, the member forfeits their ability to participate as an Active Member in the following year.
 \\* \\*
 A member may be given a conditional to complete as a means of making up for missing requirements.
-A conditional may be proposed by any member present at the Membership Eval.
+A conditional may be proposed by any member present at the Eval.
 If the conditional is approved by the Evals Director, it is then voted on by the attending members as described in \ref{Membership Evaluation Voting}.
 Each conditional consists of a set of additional requirements and a deadline to complete them.
-Members who are given a conditional have their Membership Eval decision deferred and therefore maintain their current membership status until the conditional is resolved.
+Members who are given a conditional have their Eval Process decision deferred and therefore maintain their current membership status until the conditional is resolved.
 When the deadline expires, the conditional will be brought before E-Board, who will assess its completeness.
-A conditional member who does not meet these additional requirements will have failed the Membership Eval.
+A conditional member who does not meet these additional requirements will have failed the Eval Process.
 
 \asubsection{Active Membership Leave of Absence}
 \renewcommand{\theenumi}{\alph{enumi}} % For this section, we want items to use letters
--- a/constitution.tex
+++ b/constitution.tex
@@ -299,35 +299,35 @@ Active Members are evaluated annually through the Membership Evaluation Process.
 
 % ACTIVE MEMBERSHIP EVALUATIONS
 \asubsubsection{Membership Evaluation}
-The Eval Process occurs once per academic year, where Active Members are evaluated.
-The Eval Process takes place during the spring semester to comply with the RIT Housing deadlines.
-The Eval Process is responsible for determining individuals who will have the option to continue Active Membership in the following academic year.
+The Membership Eval occurs once per academic year, where Active Members are evaluated.
+The Membership Eval takes place during the spring semester to comply with the RIT Housing deadlines.
+The Membership Eval is responsible for determining individuals who will have the option to continue Active Membership in the following academic year.
 This meeting is open only to Active Members.
 All Active Members are expected to attend this meeting.
 
 \asubsubsubsection{Membership Evaluation Voting}
-Any Active Member who has completed all the requirements as defined in \ref{Expectations of Active Members} at the beginning of the Eval Process pass the Eval Process without being voted on.
-All Active Members who have not received an exemption from E-Board prior to the start of the Eval Process will be evaluated on a member-by-member basis by Active Members.
-The Eval Process shall hold members to the objective requirements defined in \ref{Expectations of Active Members} and determine which members may continue as an Active Member in the following academic year.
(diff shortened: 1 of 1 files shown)


constitution.tex
```diff
@@ -295,18 +295,18 @@ \subsection{Modifying the Constitution}
 \end{enumerate}
 
 \asubsection{Active Membership Evaluations}
-Active Members are evaluated annually through the Eval Process described below.
+Active Members are evaluated annually through the Evaluation process described in the following.
```
Qelxiros: This is incorrect

constitution.tex
```diff
@@ -295,18 +295,18 @@ \subsection{Modifying the Constitution}
...
 
 % ACTIVE MEMBERSHIP EVALUATIONS
 \asubsubsection{Membership Evaluation}
 The Membership Eval occurs once per academic year, where Active Members are evaluated.
-It is performed as part of the Eval Process that takes place during the spring semester to comply with the RIT Housing deadlines.
+It is performed as part of the Eval process that takes place during the spring semester to comply with the RIT Housing deadlines.
```
Qelxiros: This is also incorrect, and inconsistent with the above

constitution.tex
```diff
@@ -317,15 +317,15 @@ \subsection{Modifying the Constitution}
...
 If the member fails, but has previously passed a Membership Eval, they will move to Alumni Membership in bad standing, as defined in \ref{Alumni Membership Selection}, at the end of the Standard Operating Session.
 In either case, the member forfeits their ability to participate as an Active Member in the following year.
 \\* \\*
 A member may be given a conditional to complete as a means of making up for missing requirements.
 A conditional may be proposed by any member present at the Eval.
-If the conditional is approved by the Evals Director, it is then voted on by the attending members as described in \ref{Voting}.
-Each conditional consists of a set of additional requirements and a deadline for completing them.
-Members who are given a conditional have their Eval decision deferred, and therefore maintain their current membership status until the conditional is resolved.
+If the conditional is approved by the Evals Director, it is then voted on by the attending members as described in \ref{Membership Evaluation Voting}.
+Each conditional consists of a set of additional requirements and a deadline to complete them.
```
Qelxiros: "for completing" was better

Qelxiros (dismissed):
Thanks!

constitution.tex
tallen42: I have ensured that "(the) Membership Eval" is the term used consistently.

constitution.tex
tallen42: I have ensured that "(the) Membership Eval" is the term used consistently.

Qelxiros:
@tallen42 "Eval Process" was fine, and you seem to misunderstand the system of abbreviations. All the abbreviations are available in Section 1.A, and they are used throughout the document, except in section headers. Please don't try to introduce your own abbreviations.

tallen42:
@Qelxiros Is it "Eval Process", "Eval process", or "Eval"? While I see that "Eval(s)" is an Abbreviation, and can reasonably be assumed that within the subsection "Membership Evaluation", "Eval" would refer to Membership Evaluation, whether process should exist or be capitalized is unclear. 

My changes were being made for the sake of clarity and consistency, but once we decide on this, I will implement this consistently (and without any new abbreviations).

Qelxiros:
"Eval Process" is correct when referring to ~~spring evals~~ the evaluation process. ~~I think~~ sometimes we just say "Eval" when referring to the event (spring evals) itself ~~, but I'm not 100% sure~~.

tallen42:
@Qelxiros Eval Process is set everywhere now, we should be good to go!

Qelxiros:
See above edits; sorry that I wasn't clear. Most instances of `Eval` or `Membership Eval` were correct.

tallen42:
My understanding from the section "Evaluations Processes" combined with all the Evals I've been in is that the Evals Director reads out the section of the constitution at the start of a Membership Evaluation, which would make it the Membership Evaluations Process, no?

constitution.tex
```diff
@@ -261,10 +261,10 @@ \subsection{Modifying the Constitution}
 Voting Members are expected to meet the requirements defined in \ref{Expectations of Voting Members}.
 
 \asubsubsection{Expectations of Active Members}
-Active Members are required to pay dues as stated in \ref{Collection of Dues}, attend all House Meetings and E-Board candidate speeches, and attend at least fifteen of the directorship meetings for each semester in which they are an Active Member.
+Active Members are required to pay dues as stated in \ref{Collection of Dues}, attend all House Meetings and E-Board candidate speeches, and attend at least fifteen directorship meetings for each semester in which they are an Active Member.
 They are also expected to sign a copy of the Membership Agreement.
 \\* \\*
-Active Members must participate in at least one major project during the academic year.
+Active Members must participate in at least one major project.
```
shaeespring, line 267: This would be a semantic change. I don't think it's what you  mean. Did you mean to change it to Standard Operating Session?

shaeespring (commented):
I'm confused as to why "Eval Process" is better. How does it distinguish itself from Introductory Eval Process without having "Membership" there?

constitution.tex
shaeespring: or "annually" perhaps. The way it is now could be interpretted is that you must participate in at least one major project throughout your active membership

shaeespring: Either way it'd be a semantic change

tallen42: Is there a difference between academic year and standard operating session?

tallen42: But also section 3.A states "Unless explicitly stated otherwise, the requirements and expectations defined in the Constitution are for the Standard Operating Session"

shaeespring: Good catch, I stand corrected.

shaeespring: Technically academic year is "defined" (but not really because there's not a real definition of academic year) by RIT to include breaks and finals weeks, whereas standard operating session does not, but in practice there's no difference

shaeespring: To my other question, I might need a grammar freak like @Qelxiros to help me understand why "Evals process" is strictly referring to Spring Evals instead of either Intro or Spring Evals

tallen42: The above comment from him:
> "Eval Process" is correct when referring to ~spring evals~ the evaluation process. ~I think~ sometimes we just say "Eval" when referring to the event (spring evals) itself ~, but I'm not 100% sure~.

And then I had a response right below it:
> My understanding from the section "Evaluations Processes" combined with all the Evals I've been in is that the Evals Director reads out the section of the constitution at the start of a Membership Evaluation, which would make it the Membership Evaluations Process, no?

tallen42: This wasn't yet resolved so I'm not sure.

shaeespring: I saw that message, but I don't understand the thought process. Sometimes when we say Eval when referring to Intro Evals as well. I was wondering if that thought came from the constitution somehow.

tallen42: All of these should probably be changed to Membership Eval, if we want to abbreviate Evaluation. I don't think we should cut "Membership" off. It is "implied" that "Eval" in a definition of a specific [name] Evaluation refers back to it, but we should probably keep it explicit.

tallen42: Do we want this to be "Membership Eval" (which uses the defined Abbreviation) or "Membership Evaluation"?

shaeespring: Whichever you decide. They're both the same. You can set the precedent for the future

shaeespring (dismissed):
I think a more pedantic person should also approve but I can't see any issues. The lack of statement stating Expectations of Active Members are yearly unless stated otherwise makes me sad, but it is correct

constitution.tex
```diff
@@ -295,19 +295,19 @@ \subsection{Modifying the Constitution}
 \end{enumerate}
 
 \asubsection{Active Membership Evaluations}
-Active Members are evaluated annually through the Eval Process described below.
+Active Members are evaluated annually through the Membership Evaluation Process.
```
Qelxiros:
```suggestion
Active Members are evaluated annually through the Membership Eval Process.
```

constitution.tex
```diff
@@ -295,19 +295,19 @@ \subsection{Modifying the Constitution}
...
 
 % ACTIVE MEMBERSHIP EVALUATIONS
 \asubsubsection{Membership Evaluation}
 The Membership Eval occurs once per academic year, where Active Members are evaluated.
-It is performed as part of the Eval Process that takes place during the spring semester to comply with the RIT Housing deadlines.
+The Membership Eval takes place during the spring semester to comply with the RIT Housing deadlines.
```
Qelxiros: You can add `Membership` here, but the previous sentence was better otherwise.

constitution.tex
```diff
@@ -295,19 +295,19 @@ \subsection{Modifying the Constitution}
...
 The Membership Eval is responsible for determining individuals who will have the option to continue Active Membership in the following academic year.
 This meeting is open only to Active Members.
 All Active Members are expected to attend this meeting.
 
-\asubsubsubsection{Voting}
-Any Active Member who has completed all of the requirements as defined in \ref{Expectations of Active Members} at the beginning of the Membership Eval passes their Membership Eval without being voted on.
-All Active Members who have not received an exemption from E-Board prior to the Membership Eval will be evaluated on a member-by-member basis by Active Members.
+\asubsubsubsection{Membership Evaluation Voting}
+Any Active Member who has completed all the requirements as defined in \ref{Expectations of Active Members} at the beginning of the Membership Eval pass without being voted on.
```
Qelxiros:
```suggestion
Any Active Member who has completed all the requirements as defined in \ref{Expectations of Active Members} at the beginning of the Membership Eval passes without being voted on.
```

constitution.tex
Qelxiros: "for completing" is still the better choice here, not sure why you resolved this.

Qelxiros (changes requested):
See comments below. Also, does the Intro Eval section need to change now to specify which Eval it is?

constitution.tex
tallen42: It feels like a matter of opinion- both read the same.
Grammarly recommends "to complete".

constitution.tex
tallen42: I strongly disagree; see the sentences before and after it. They both start with "The Membership Eval" - it seems weird that the sentence in the middle was "It is performed as part of the Eval Process"
Also, the Eval Process is a weird term that really only needs to apply for 6 weeks since that's a process. The only "Process" for active membership is passing the Membership Evaluation in the spring. Calling this a Process is weird and feels highly incorrect

constitution.tex
Qelxiros: 1. That would have been good to mention a month ago, instead of resolving the comment without a reply.
2. If you want to use the infinitive, the full sentence should be `Each conditional consists of a set of additional requirements and a deadline by which to complete them.`, but even then it's better to say `...by which they must be completed.` Both of those are clunkier than what was already here.
3. Grammarly should never be the source of truth.

constitution.tex
tallen42: I get why you want this, it's a defined abbreviation, and then outside of the section titles, we should use abbreviations. However, in the context of the document, I think it makes more sense to have it say the title of the thing right below it. 

I am happy to hear other opinions on this though

constitution.tex
tallen42: I will revert it to "for completing" but this one makes not much sense to me I will be honest

tallen42: However, we use "to complete" in every other reference of conditionals

constitution.tex
Qelxiros: The established precedent wrt abbreviations is to exclusively use the short form outside of headings. This is (afaik) consistent throughout the constitution. I don't see a good reason to break that consistency here. You say it makes more sense one way; I say it makes more sense the other way. This is why we have precedent.

constitution.tex
tallen42: I see how it's different though

constitution.tex
Qelxiros: The paragraph flows better if we allow ourselves to use pronouns. Three sentences in a row that start with `The Membership Eval` is stuffy. 

If you really want to get rid of the Process reference, I can get behind that.

tallen42: I added pronouns

Qelxiros (approved):

PR #289, opened 2026-03-30
289 making titles / substitle have consistent spacing

ikopke23:
subsections now all have a newline after them when appropriate using titlesec 

added title format title spacing to \asubsection \asubsubsection \asubsubsubsection

Check one:

Summary of change(s):

My change
image

Live render
image

Note the urls for confirmation

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -41,17 +41,23 @@
 \renewcommand{\thesubsubsection}{\arabic{section}.\Alph{subsection}.\arabic{subsubsection}}
 \titleformat{\section}{\normalfont\Large\bfseries}{Article \thesection}{1em}{}
 \titleformat{\subsection}{\normalfont\large\bfseries}{Section \thesubsection}{1em}{}
+\titleformat{\subsubsection}{\normalfont\normalsize\bfseries}{\thesubsubsection}{1em}{}
+\titlespacing*{\subsubsection}{0pt}{3.25ex plus 1ex minus .2ex}{0.5em}
 
 % Adding an \asubsubsection -- I feel dirty
 %\setcounter{secnumdepth}{5}
 \newcommand{\asubsubsection}[1]{\paragraph{#1} \label{#1}}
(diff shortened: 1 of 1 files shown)


shaeespring (changes requested):
Ah!!! This looks so good! I have one qualm, and that's in section 5.A.4 and there's probably other sections that do this, but now there's weird indentation

Normal: 
image
But with the change:

image


It's weird because I don't see why your change would affect indentation like this. Do you have any ideas on how to fix it?

shaeespring:
There are multiple other sections that do this now that I noticed it

ikopke23:
> Ah!!! This looks so good! I have one qualm, and that's in section 5.A.4 and there's probably other sections that do this, but now there's weird indentation
> 
> Normal: image But with the change:
> image
> 
> It's weird because I don't see why your change would affect indentation like this. Do you have any ideas on how to fix it?

Its due to the fact that people are putting newlines in things, it denotes a new paragraph which then means the new paragraph should be indented as per usual, however we tend to not want that so I am reviewing the times where this happens to see if they should be separate or just one paragraph

shaeespring:
I think the issue still occurs in 4.B.4 (at the end), 5.A.4.A (in the middle), and 6.A.3 (After the table). This looks super awesome and is really great. I'm sorry for being so picky. I can make the changes if you like (since you figured out what it was. Massive W)

constitution.tex
```suggestion
```

```suggestion
```

```suggestion
```

shaeespring (changes requested):
I think these lines are the only issues. I'll double check again too

ikopke23:
> I think the issue still occurs in 4.B.4 (at the end), 5.A.4.A (in the middle), and 6.A.3 (After the table). This looks super awesome and is really great. I'm sorry for being so picky. I can make the changes if you like (since you figured out what it was. Massive W)

You're all good, I somehow missed that you replied my bad, will get this committed, the web ui is not letting me for some reason

shaeespring (approved):
THANK YOU! This looks so much better imo

PR #295, opened 2026-04-18
fix: Proper section depth for RTPs

tallen42:
Check one:

- [x] I don't know. I see it as changing bullet points (non-semantic), but maybe it is semantic? 

Summary of change(s):

Corrected the section depth for subsections within RTP definition.


shaeespring (approved):
Good catch!

PR #296, opened 2026-04-19
Add CI workflow

shaeespring:
Check one:

Summary of change(s):
Makes it so reviewers can focus on semantics, not silly other things, on their local machine

~Could also do the Docker thing, but it's the same thing and takes so much longer.~

I forgot make won't work if it has nothing else installed :/

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,17 @@
+name: Makefile CI
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


shaeespring:
Okay, I think this is what I want. Basically, if your code doesn't pass the check, it's probably not ready for review

Qelxiros (approved):
Great, thanks!

PR #297, opened 2026-04-20
Only allow Voting Members to sign impeachment petitions

pikachu0542:
Check one:

Summary of change(s):
Fix loophole, since Alumni are "current members"

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -699,13 +699,13 @@ The vote of a vacated E-Board position shall be cast as an abstention in all E-B
 
 \asubsubsection{Impeachment}
 \begin{enumerate}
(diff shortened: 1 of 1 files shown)


pikachu0542:
Updated the new phrasing to say "Active Members" instead of "Voting Members", because discussions of this topic have occurred, and people think that people who are active but dont meet gatekeep should still be able to sign an impeachment petition

shaeespring:
It's not a loophole imo, and probably shouldn't be advertised as such. Currently Alumni can sign impeachment petitions, and that's just a fact and a feature, not a loophole

Grammatically correct PR, but the branch name and summary are misleading imo

shaeespring:
Will approve when branch name and summary are fixed

PR #299, opened 2026-05-01
Removed quorum for cloud spending presentation votes

DeadPixelPro:
We do not track quorum for cloud expenditures.

Check one:

Summary of change(s):
Removed the fifty percent quorum requirement for spending presentations that pull from "the cloud." We do not currently track quorum for this, and no other spending presentation requires a quorum.

(spotted by @pikachu0542)


pikachu0542 (approved):
lgtm

pikachu0542:
Just for reference, spending presentations from a directorship budget do not require a quorum

DeadPixelPro:
While I'm here, I should probably add something. All this currently does is make this more in line with how things currently go, but considering that a lot can be done with an expenditure like this, I think it would make a lot of sense if we instead modify this so that cloud expenditures _above a certain amount_ require a quorum.

PR #308, opened 2026-09-01
Rearranged eboard positions

Titanium-Lung:
Check one:

Summary of change(s):
Rearranged eboard positions to be: chair, evals, financials, socials, rnd, imps, opcomm, history, PR.

### DIFF
--- a/constitution.tex
+++ b/constitution.tex
@@ -836,9 +836,9 @@ After this date, dues collection is suspended until the start of the next academ
 		\hline
 		OpComm       & 20\%
 		\\ \hline
-		R\&D         & 20\%
-		\\ \hline
 		Social       & 20\%
+		\\ \hline
+		R\&D         & 20\%
 		\\ \hline
 		IMPs         & 15\%
 		\\ \hline
--- a/constitution.tex
+++ b/constitution.tex
@@ -508,11 +508,21 @@ A Voting E-Board Member is considered to be any E-Board Member with a non-zero v
 	\item To vote on issues presented to E-Board
 \end{enumerate}
 
-\asubsubsection{Responsibilities of the Operational Communications Director}
+\asubsubsection{Responsibilities of the Socials Director(s)}
 \begin{enumerate}
-	\item To represent the RTPs at E-Board Meetings and at House Meetings
-	\item To report the status of the house systems, services, and networks
-	\item To facilitate communications with RIT Information and Technology Services
+	\item To exercise general supervision over Social operations
+	\item To oversee the organization and execution of CSH social activities
+	\item To ensure that there is a variety of events for members to participate in throughout the academic year
+	\item To vote on issues presented to E-Board
+\end{enumerate}
+
+\asubsubsection{Responsibilities of the Public Relations Director}
+\begin{enumerate}
+	\item To operate CSH social media accounts intended to represent CSH as a whole
+	\item To oversee the organization and execution of CSH philanthropic events
+	\item To preserve and improve the public image of CSH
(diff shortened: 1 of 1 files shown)


pikachu0542 (dismissed):
lgtm

shaeespring (changes requested):
If the goal is consistency here, abbreviations and breakdown of dues also need reordered

shaeespring (changes requested):
For all except Financials, I believe they should be moved to their previous positions for historical reasons. They've been in this order since the constitution was a Word document, and I think that's pretty cool. Since no deliberate reason was presented for changing it (other than financials - which is the change I agree with) I don't agree with the change. If a reason can be presented I will reevaluate

constitution.tex
```diff
@@ -587,7 +587,7 @@
 	\item Candidates must be enrolled students at RIT, must be in good academic standing with both RIT and their academic department, and may not be on a conduct sanction of disciplinary probation or higher during the term of office.
 \end{enumerate}
 
-\asubsubsection{Qualifications for Chairperson, Evaluations Director, Social Director, Financial Director, Research and Development Director(s), House Improvements Director, History Director, Public Relations Director}
+\asubsubsection{Qualifications for Chairperson, Evaluations Director, Financial Director, Research and Development Director(s), Social Director(s), House Improvements Director, History Director, Public Relations Director}
```
gambogi:
```suggestion
\asubsubsection{Qualifications for Chairperson and Directors}
```

Seems like this section would always apply to all positions

constitution.tex
```diff
@@ -631,7 +631,7 @@
 	      When attendance requirements call for the dual directorship position to be present, at least one of the dual directors must be present to fulfill the requirements.
 \end{enumerate}
 
-\asubsubsection{Selection of the Chairperson, Evaluations Director, Social Director(s), Financial Director, House Improvements Director, History Director,  Research and Development Director(s), Public Relations Director}
+\asubsubsection{Selection of the Chairperson, Evaluations Director, Financial Director, Research and Development Director(s), Social Director(s), House Improvements Director, History Director, Public Relations Director}
```
gambogi: Ditto here

gambogi (commented):
Randomly checked in and I was pleasantly surprised this is still actively maintained! 
For what it’s worth, I don’t really think the order of these individual sections matters too much. It does affect any references to section numbering/ref but that’s fine. 
Take my suggestion about the catchall headings if you like. I’ll butt out, do what feels best :)

constitution.tex
Qelxiros: opcomm is different here and below unfortunately

constitution.tex
```diff
@@ -587,7 +587,7 @@
 	\item Candidates must be enrolled students at RIT, must be in good academic standing with both RIT and their academic department, and may not be on a conduct sanction of disciplinary probation or higher during the term of office.
 \end{enumerate}
 
-\asubsubsection{Qualifications for Chairperson, Evaluations Director, Social Director, Financial Director, Research and Development Director(s), House Improvements Director, History Director, Public Relations Director}
+\asubsubsection{Qualifications for Chairperson, Evaluations Director, Financial Director, Social Director, Research and Development Director(s), House Improvements Director, History Director, Public Relations Director}
```
```suggestion
\asubsubsection{Qualifications for Chairperson, Evaluations Director, Financial Director, Social Director(s), Research and Development Director(s), House Improvements Director, History Director, Public Relations Director}
```

shaeespring (dismissed):
Sorry. I haven't been getting a whole lot of time to do anything. I think you had this change at one point, but I like that you did it in the other place. Not requesting changes because it's not your fault that it is the way that it is. I'll hit the button if you say so

shaeespring (approved):
Yay!

